PR #13, opened 2014-06-25
Added #hideme tag & shift + plus operator options

arulalant:
I added two options

NOTE_8: Use shift+Plus operator (keys shit+) to go to or jump into next code cell by showing all markup fragments which are all presented in between two code cells.

NOTE_9: Use `#hideme` tag inside code cell as starting line to hide code cell or [In] alone, but it shows its correspoding outputs [Out] only. This will work when you click on live slide / presentation mode. This is useful when you load image by using Image. For eg,

#hideme
from IPython.core.display import Image 
Image(filename='img/hello-python.png') 

The above code cell will show only its output image during the live presentation and it will not show python code. Once you come out of live presentation mode, then all `#hideme` code will be visible in nbviewer.
PS : `#hideme` tag works only in firefox browser. For chrome, need to test/work out.

Please look at the file changed option to view diff.

Thanks.


damianavila:
I will close this one for now because it does not merge cleanly and I pretend to discuss other option in #32. So, it makes no sense to left it open until we agreed in something... Thanks @arulalant and sorry to no be enough responsive here... This is important even if it is not merged because it provides a mod for people interested in your solution. Let's continue the discussion in the issue to see where we arrive. Cheers.

PR #40, opened 2014-11-02
allow use of leap motion

Carreau:
Kind of buggy, and sometime hard to change slides, but it works.


damianavila:
I need a leap motion to test this out!!! This is sometimes configurable, so I will merge it because it seems to not affect other things and give a nice feature for people to play with. Thanks @Carreau!
Feel free to submit PR fixing the buggy things :stuck_out_tongue_winking_eye:

PR #50, opened 2015-01-26
Make reveal work with new notebook layout

jhamrick:
This makes changes so that the following notebook:

image

now looks like this when using the live reveal plugin on IPython master:

image

### DIFF
--- a/livereveal/main.js
+++ b/livereveal/main.js
@@ -125,14 +125,14 @@ function labelIntraSlides(){
 
 function Slider(begin, end, container) {
   // Hiding header and the toolbar
-  $('div#header-container').toggle();
+  $('div#header').toggle();
   $('div#maintoolbar').toggle();
   IPython.menubar._size_header();
   //if(event)event.preventDefault();
 
   // switch the panel back color to white (it does not work in css,
   // I do not why, so switching by js)
-  $('div#notebook_panel').css('background-color','#ffffff');
+  $('div#site').css('background-color','#ffffff');
 
   /*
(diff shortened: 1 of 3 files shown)


livereveal/main.js
```diff
@@ -348,15 +351,18 @@ function buttonExit() {
 }
 
 function Remover() {
-  $('div#notebook_panel').css('background-color','#ffffff');
-  $('div#header-container').toggle();
+  $('div#site').css("height", "");
```
damianavila: I see you use this `""` (empty) definitions with the css... I am not a css guy, so... just asking before looking into the net... this "practice" allows you to delete the current css?

jhamrick: Yep. To be honest, I have only tested this in Chrome so far, but I'm pretty sure it works in other browsers too.

damianavila: For completeness... 
```
Setting the value of a style property to an empty string — e.g. $( "#mydiv" ).css( "color", "" ) — 
removes that property from an element if it has already been directly applied, whether in the 
HTML style attribute, through jQuery's .css() method, or through direct DOM manipulation of the 
style property. It does not, however, remove a style that has been applied with a CSS rule in a 
stylesheet or <style> element. Warning: one notable exception is that, for IE 8 and below, 
removing a shorthand property such as border or background will remove that style entirely from 
the element, regardless of what is set in a stylesheet or <style> element.
```

This is nice... thanks @jhamrick for let me know...

damianavila:
Thanks @jhamrick

PR #57, opened 2015-02-05
Ubuntu 14.04 IPython notebook does not have IPython.menu._size_header

tjwei:
live_real did not work for me. My IPython notebook is ver 2.31 bundled in Ubuntu 14.04 (apt-get version).
I check the javascript console and find out an error caused by `IPython.menu._size_header` not defined. After I commented out these line, live_real seems to work fine. 
So I add a check to see whether `IPython.menu._size_header` is defined before calling it.

### DIFF
--- a/livereveal/main.js
+++ b/livereveal/main.js
@@ -127,7 +127,9 @@ function Slider(begin, end, container) {
   // Hiding header and the toolbar
   $('div#header').toggle();
   $('div#maintoolbar').toggle();
-  IPython.menubar._size_header();
+  if(IPython.menubar._size_header) {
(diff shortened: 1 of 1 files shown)


damianavila:
WAT? it is working normally only with this change? I would expect master do not work with 2.3.1, since of the layout machinery change a lot....
OK, I will take a look at this and test it and if it fully working, I am surely merging this soon... thanks!

damianavila:
This fix just one of the issues, I will submit another PR solving the rest of them, so master will be compatible again with 2.3.x and 2.4.x versions of IPython... merging this one as is. Thanks!

Issue #29, opened 2014-07-24
Enforce (or not) slide at the first cell


damianavila:
From #26:

I'm running into issues with trying to "skip" cells in slideshows with live_reveal. First, if you try to skip the very first fragment (e.g. if it contains import statements, etc.), the cell is not skipped, and the state of the cell is switched to "slide" from "skip" automatically.
This is a design decision because a lot of people could forget to setup the first cell to show as an slide... but you are right about some uses cases where is good to hide the first (or more) cells... I think I have to re-evaluate this decision or come up with a better solution...

damianavila:
Just to give an update, I still think the same about this, targeting to 3.x compatibility.

takluyver:
FWIW, I just ran into this and had a WTF moment - I wanted some imports at the top of the notebook so I'd remember to run them before going into presentation mode, but they would be hidden in presentation mode.

Though I can also see that it would be surprising if you went into slide mode and were missing content unexpectedly.

Perhaps automatically make the first cell 'slide' if it doesn't have a slide type set, but if the user sets the slide type to 'skip', assume they mean it.

damianavila:
>Perhaps automatically make the first cell 'slide' if it doesn't have a slide type set, but if the user sets the slide type to 'skip', assume they mean it.

Seems a good compromise...

PR #59, opened 2015-02-12
Allow cells at beginning and end to be notes/skip

takluyver:
This keeps the special handling for the first and last cells, but now it finds the first and last *visible* cell (visible = not skip or notes), so you can have a skipped cell at the beginning.

Closes gh-29

### DIFF
--- a/livereveal/main.js
+++ b/livereveal/main.js
@@ -93,8 +93,18 @@ function labelCells(){
     }
   }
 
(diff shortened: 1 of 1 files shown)


takluyver:
I fixed the issue I introduced with skipped slides being rearranged; I thought it was necessary to find the last visible slide as well, but seemingly it's not.

damianavila:
LGTM, merging...

Issue #58, opened 2015-02-12
note cells stay hidden after leaving presentation mode


takluyver:
If I set some cells as notes using the cell toolbar, they are hidden when I switch into presentation mode (presumably usable with a speaker view), but they don't come back when I leave presentation mode. Refreshing the page restores them.

takluyver:
Also, and possibly related, cells marked 'skip' seem to come back at the top of the file regardless of where they were before.

takluyver:
OK, the problem with skipped cells is something I've introduced in the PR I just opened. The problem with note cells is in master, however.

damianavila:
OK, I will take a look at the PR...

PR #60, opened 2015-02-12
Bring notes back when leaving slideshow mode

takluyver:
Instead of forcefully hiding note cells, add them to a class that's hidden during slideshow mode.

Closes gh-58

### DIFF
--- a/livereveal/main.css
+++ b/livereveal/main.css
@@ -64,6 +64,10 @@
   display: none !important;
 }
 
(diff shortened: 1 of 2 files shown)


damianavila:
LGTM too, sometimes I like to change the css on the fly because some incompatibilities between the notebook css and reveal css... in this case, the class seems to work OK, so merging...

PR #65, opened 2015-02-15
WIP - Rewrite code to build reveal.js HTML structure

takluyver:
Gets rid of metadata.internals, (slide|subslide|fragment)_helper, and several loops.

There's still some work to do - I think the behaviour of fragments is probably wrong at present. I will look into this this afternoon.

### DIFF
--- a/livereveal/main.js
+++ b/livereveal/main.js
@@ -34,177 +34,69 @@ if(IPython.version.substring(0, 1) === '2') {
     console.log("IPython version: 3.x");
 }
 
-function setupDict(){
-  var cells = IPython.notebook.get_cells();
-  for(var i in cells){
-    var cell = cells[i];
-    if (cell.metadata.slideshow === undefined){
-      cell.metadata.slideshow = {};
-      cell.metadata.slideshow.slide_type = '-';
-    }
-    if (cell.metadata.internals === undefined){
-      cell.metadata.internals = {};
-      cell.metadata.internals.slide_type = '-';
-    }
-  }
-}
+/* Use the slideshow metadata to rearrange cell DOM elements into the
+ * structure expected by reveal.js
+ */
+function markupSlides(container) {
+    var slide_section = $('<section>').appendTo(container);
+    var subslide_section = $('<section>').appendTo(slide_section);
 
-function labelCells(){
-  var cells = IPython.notebook.get_cells();
-  for(var i=1; i< cells.length; i++){
-    var cell = cells[i];
-    var index = i - 1;
-    if (cell.metadata.slideshow.slide_type == 'slide') {
-      cells[i].metadata.internals.slide_type = 'slide';
-      cells[index].metadata.slide_helper = 'slide_end';
-      cells[index].metadata.internals.slide_helper = 'slide_end';
-    }
-    else if (cell.metadata.slideshow.slide_type == 'subslide'){
-      cells[i].metadata.internals.slide_type = 'subslide';
-      cells[index].metadata.slide_helper = 'subslide_end';
-      cells[index].metadata.internals.slide_helper = 'subslide_end';
-    }
-    else if (cell.metadata.slideshow.slide_type == 'fragment'){
-      cells[i].metadata.internals.slide_type = undefined;
-      cells[index].metadata.slide_helper = undefined;
-      cells[index].metadata.internals.slide_helper = undefined;
-      cells[i].metadata.internals.frag_number = i;
-      var j=1;
-      while (j < cells.length - i) {
-        cells[i + j].metadata.internals.frag_helper = 'fragment_end';
-        cells[i + j].metadata.internals.frag_number = i;
-        j++;
-      }
-    }
-    else if (cell.metadata.slideshow.slide_type == 'notes'){
-      cells[i].metadata.internals.slide_type = undefined;
-      cells[index].metadata.slide_helper = undefined;
-      cells[index].metadata.internals.slide_helper = undefined;
-    }
-    else if (cell.metadata.slideshow.slide_type == 'skip'){
-      cells[i].metadata.internals.slide_type = undefined;
-      cells[index].metadata.slide_helper = undefined;
-      cells[index].metadata.internals.slide_helper = undefined;
-    }
-    else if (cell.metadata.slideshow.slide_type == '-'){
-      cells[i].metadata.internals.slide_type = undefined;
-      cells[index].metadata.slide_helper = undefined;
-      cells[index].metadata.internals.slide_helper = undefined;
-    }
-  }
-
-  // The first cell that isn't explicitly skipped or notes needs to be the
-  // start of a slide.
-  var slide_type;
-  for (i=0; i < cells.length; i++) {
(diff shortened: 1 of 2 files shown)


damianavila:
This is looking great so far... as I said previously, any improvement to make this part of the code understandable for everyone is very welcome... I will wait for you on the fragment thing to make a full review, but at first sight, it is a nice improvement! After you PR I plan the make other clean up to simplify things... Thanks for working on this...

damianavila:
Btw, I like the logic used here because it is a similar to the logic used in the static slides... and get rid of those hacky loops...

takluyver:
I'm also seeing something strange - when I go into slideshow mode, it often starts out blank except for the controls, help button, etc., and the arrow keys don't do anything. Then when I click on the controls to go forward or backward a slide, it starts working as expected - slides show, and the left/right keys work. Have you seen this before? I'm not quite sure how to debug it?

damianavila:
>Have you seen this before? 

I saw it, but not frequently... btw, this is something getting worse with this PR, or in general?

>I'm not quite sure how to debug it?

Probably something related with reveal initialization... but not sure, I have to explore it... btw, some things are usually difficult to debug in this extension, at least in my experience... jeje

takluyver:
Fragments are fixed.

I've worked out what's causing the error I'm seeing - it's some interaction between browser history and reveal.js interaction. This had actually confused me before - it seemed to start the slideshow on semi-random slides, it's only now I realise that it's going back to the slide (number) that was last used in slideshow mode.

I can fix this either by restoring the HTML IDs on the slides, or by setting `history: false` to turn off those features of reveal.

I'm actually leaning towards just disabling the history:
- The IDs don't work as links to the slides, because the notebook doesn't open in slideshow mode.
- Even if I fix it, you'll probably still get weird results if you view through to the last slide, then switch out of slideshow mode, cut out one slide and go back to slideshow mode - because it will be trying to go to e.g. slide 9 out of 8.
- As I found, it's not obvious that it's restarting the slideshow at the point where you last left it. The most common behaviours in presentation programs I know of are to start a slideshow from the beginning each time, or from the selected slide (the analogy here would be the cell containing the selected slide.

takluyver:
Rats, no, that's not it, or at least not the whole story.

damianavila:
I agree on disabling `history` for the reasons you exposed before... but I also think we need to provide this:

>from the selected slide (the analogy here would be the cell containing the selected slide.

And probably it is easy to do... Can you open a new issue for that?

> Rats, no, that's not it, or at least not the whole story.

So, you are still seeing the behaviour...

takluyver:
Yep, I'm still seeing the behaviour. It works OK the first time I launch the slideshow after loading the page, but if I leave slideshow mode and come back, the problem appears. This is something that my changes have introduced - if I go back to your most recent commit, it is OK.

Will open a new issue for slideshow-from-selected-cell.

takluyver:
I can break it down into two separate issues: the part where the left/right keys on the keyboard don't work is just a focus issue when I click on the toolbar button to start reveal. Starting the slideshow with Alt+R doesn't have that bug.

The more confusing bug, which I see with both methods of starting it, is that the first cell is 'invisible' when I go into slideshow mode the second (or later) time. I can flip through fragments in it and see the controls updating, but the slide itself is invisible until I move to another slide, which is displayed normally (and then I can change back to the first one and it is fine).

takluyver:
It looks like each subsequent call to Revealjs.initialize() is creating another instance - the second time I go to slideshow mode, the 'ready' event is firing twice, the third time it fires three times. Not sure if this is related, but it seems like that's probably not ideal.

takluyver:
Nope, that's not the cause - I see the same pattern with that event firing multiple times on your most recent commit, which doesn't exhibit the problem.

takluyver:
OK, in fact it *was* the browser history thing, but turning the reveal history feature off didn't fix it. The only way I could find to get it work correctly was to restore the slide IDs and leave the history feature turned on in reveal.js. So for now it resumes the slideshow from where you left it off, based on the URL hash.

If we want to launch slideshows from the beginning by default, it should just be a matter of setting the URL hash to `#slide-0-0` before we initialise reveal. But I'll leave that for another PR.

I believe this is now working if you want to review it more carefully.

livereveal/main.js
```diff
@@ -34,177 +34,88 @@ if(IPython.version.substring(0, 1) === '2') {
...
-
-}
-
-function labelIntraSlides(){
-  var cells = IPython.notebook.get_cells();
-  for(var i in cells){
-    var cell = cells[i];
-    if (cell.metadata.slideshow.slide_type == 'fragment') {
-      $('.cell:nth('+i+')').addClass("fragment");
-      $('.cell:nth('+i+')').attr('data-fragment-index', cell.metadata.internals.frag_number);
```
damianavila: I will look all the PR soon, but with a really quick view I don't see the `frag_number` anywhere in your refactor, it is important to keep it (I mean include the concept in the refactor) to force the linearity in the fragments "apparition"...

takluyver: I don't follow what you mean by 'linearity'. As I understand it, if you don't add a `data-fragment-number`, they'll just appear in the order they are in the DOM, which should be correct.

damianavila: >which should be correct.

But that's not correct if you want to keep the flow of information in a "linear" way... This is the relevant discussion: https://github.com/ipython/ipython/pull/4356 and how I solved for the static slides: https://github.com/ipython/ipython/pull/5957, which is essentially what I did here too.

takluyver: Ah, gotcha. I think I've achieved the same thing in a different way. Given a 'fragment' cell followed by a '-' cell, your code was making something like this:

```html
<div class="fragment" data-fragment-index="1">cell 1</div>
<div class="fragment" data-fragment-index="1">cell 2</div>
```

My code produces something like this:

```html
<div class='fragment'>
  <div>cell1</div>
  <div>cell2</div>
</div>
```

i.e. it nests the cells of a single fragment in another container so they appear together. This was what I meant when I said that I needed to do something more on fragments - you can see the change I made to fix it [here](https://github.com/takluyver/live_reveal/commit/65a33bc37c0c00f32c09241610ec1ed5c3781557).

damianavila: yep, I did not see that because it was just a quick view :wink: (and I was thinking in my solution)... btw, nice way to solve it... 
OK, I will do the full review tomorrow, getting late here, but I think we will merge this one very soon.

damianavila:
BTW, are you still seeing this in this branch?

>the part where the left/right keys on the keyboard don't work is just a focus issue when I click on the toolbar button to start reveal. Starting the slideshow with Alt+R doesn't have that bug.

If that's the case, please open an issue to deal with that...

takluyver:
Oh yes, I'd forgotten about that one. It's fixed now.

damianavila:
Finally, I have time to test it...
Code look clean and now more easy to reach for others... 
Thanks for working on this, let's keep making RISE even better :wink:
Sorry for the delay and thanks again!!
Merging now...

takluyver:
Thanks Damián!

Issue #55, opened 2015-02-01
Revisit configuration options


damianavila:
Better way to pass a config object...

takluyver:
I am happy to work with you on using the new frontend config system for this. That will make it require IPython 3, but looking at #54 I think that's your plan anyway.

damianavila:
Yes, I would love to see this solved by the new frontend config system... I have no problem to require IPython 3, as you said is on my plans... if the user wants to use the new feature, then IPython 3 is a must...

PR #72, opened 2015-02-25
Use frontend config

takluyver:
Another significant change, but not such a big diff as the last one. This uses the new frontend config system to store user preferences, rather than putting them in custom.js where the extension is loaded.

This has the great advantage that it allows livereveal to be loaded as a regular extension, which allows the installation code to be much simpler. The new installation code will fail with an ImportError on IPython 2.x, rather than installing something that won't work.

N.B. Any users who have symlinked the extension using `setup.py install --develop` will need to reinstall it after this change, because the call to `livereveal.parameters()` in their custom.js won't work.

Addresses #55

### DIFF
--- a/livereveal/main.js
+++ b/livereveal/main.js
@@ -8,6 +8,19 @@
 * -----------------------------------------------------------------------------
 */
 
+define(['jquery',
+        'base/js/utils',
+        'services/config',
+], function($, utils, configmod) {
+
+var config_section = new configmod.ConfigSection('livereveal',
+                            {base_url: utils.get_body_data("baseUrl")});
+config_section.load();
+var config = new configmod.ConfigWithDefaults(config_section, {
+    theme: 'simple',
+    transition: 'linear'
+});
+
 IPython.notebook.get_cell_elements = function () {
   /*
   * Version of get_cell_elements that will see cell divs at any depth in the HTML tree,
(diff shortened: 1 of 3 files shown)


livereveal/main.js
```diff
@@ -8,6 +8,19 @@
...
+        'base/js/utils',
+        'services/config',
+], function($, utils, configmod) {
+
+var config_section = new configmod.ConfigSection('livereveal',
+                            {base_url: utils.get_body_data("baseUrl")});
+config_section.load();
+var config = new configmod.ConfigWithDefaults(config_section, {
+    theme: 'simple',
+    transition: 'linear'
```
damianavila: we should eventually set up others config options... but that's for another PR...

setup.py
```diff
@@ -1,68 +1,22 @@
...
-});
-"""
-
-
-def install(use_symlink=False, profile='default'):
+def install(use_symlink=False, profile='default', enable=True):
     # Install the livereveal code.
     install_nbextension(livereveal_dir, symlink=use_symlink,
-                        overwrite=use_symlink)
+                        overwrite=use_symlink, user=True)
```
damianavila: we should add the possibility of a system-wide installation too... but we can discuss that later...

damianavila:
I left two comments but those things are for later... so merging now... thanks again @takluyver, this is great!

PR #78, opened 2015-02-27
Config for where the slideshow starts

### LINKED ISSUE #66: Start slideshow from selected cell
From discussion on PR #65: it would be useful to have a feature where the slideshow can be started from the selected cell.

takluyver:
Slideshows start at the beginning by default, but by setting a config option, you can make them start at the slide containing the currently selected cell.

Closes #66

### DIFF
--- a/livereveal/main.js
+++ b/livereveal/main.js
@@ -118,7 +118,12 @@ function markupSlides(container) {
     return selected_cell_slide;
 }
 
-// Set the #slide-x-y part of the URL to control where the slideshow will start
+/* Set the #slide-x-y part of the URL to control where the slideshow will start.
+ * N.B. We do this instead of using Reveal.slide() after reveal initialises,
+ * because that leaves one slide clearly visible on screen for a moment before
+ * changing to the one we want. By changing the URL before setting up reveal,
+ * the slideshow really starts on the desired slide.
+ */
 function setStartingSlide(selected) {
     var start_slideshow = config.get_sync('start_slideshow_at');
     if (start_slideshow === 'selected') {
--- a/livereveal/main.js
+++ b/livereveal/main.js
@@ -18,7 +18,8 @@ var config_section = new configmod.ConfigSection('livereveal',
 config_section.load();
 var config = new configmod.ConfigWithDefaults(config_section, {
     theme: 'simple',
-    transition: 'linear'
+    transition: 'linear',
+    start_slideshow_at: 'beginning',
 });
 
 IPython.notebook.get_cell_elements = function () {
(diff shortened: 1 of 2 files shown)


livereveal/main.js
```diff
@@ -106,8 +115,22 @@ function markupSlides(container) {
...
     $('.end_space').appendTo('div#notebook-container');
+    return selected_cell_slide;
 }
 
+// Set the #slide-x-y part of the URL to control where the slideshow will start
+function setStartingSlide(selected) {
+    var start_slideshow = config.get_sync('start_slideshow_at');
+    if (start_slideshow === 'selected') {
+        // Start from the selected cell
+        window.location.hash = "/slide-"+selected[0]+"-"+selected[1];
```
damianavila: What do you think about using the Reveal API to do this?
https://github.com/hakimel/reveal.js#api, especially `Reveal.slide( indexh, indexv, indexf );` 
I think we should use this API as far as we can...

takluyver: Oh, I hadn't seen that. Yes, that is neater. It presumably needs to be done in the 'ready' callback, though, rather than before the slideshow starts like this.

damianavila: yep... that's was my thought... sorry to not tell you before in the issue...

takluyver: Not to worry. I may get to this at the weekend, but if you get there first, feel free to reuse any parts of this.

damianavila: I will not be available for two weeks, so go ahead when you have some time...

takluyver: I just tried doing it this way, but I didn't like the result - the last visible slide is very clearly on the screen for a moment after going into slideshow mode before it changes to the cell you want, whereas the hash-changing method in this PR really makes it start at the desired slide.

damianavila: Well... that's interesting... if the behaviour is better in this way, I would not be opposed to use it... but let's make a tiny comment about that fact in the source code...
I will try to review it soon... just arrive home today :wink:

takluyver: No problem, comment added.

damianavila:
OK, this LGTM... can you add a reference in the README, here: https://github.com/damianavila/RISE#configure-your-own-options
After that, we can merge this one immediately...

takluyver:
Done :-)

damianavila:
Amazing... sorry to bother you so much... and for the delay (I got more than 1000 issues on github when I come back from vacation and more than 350 was from IPython, so I had a of read to be updated ;-)

takluyver:
No problem - I know what it's like having to wade through hundreds of issues.

PR #83, opened 2015-03-17
Hide anchor links

adl:
When you have a slide with section names, such as

    # The title
    contents...

the notebook displays an anchor link `¶` at the end of section name, so that you can easily link between sections.

I see no point in showing this anchor link in presentation mode: it is distracting every time the mouse passes over it, and it does not even work (clicking it moves you to the start of the presentation).

How about hiding it?


damianavila:
Sound reasonable, it makes no sense in the slideshow mode, and as you said, it does not even work, so I agree with the solution... let's merge it!

damianavila:
Thanks for your contribution!!

PR #84, opened 2015-03-18
Hide the cell toolbar in presentations (Fixes #39)

### LINKED ISSUE #39: Per-Cell toolbar still active when presentation starts
There is a slight problem when presentation is shown on screen with any per-cell toolbar active ("Edit metadata", "Raw Cell" or "Slideshow") that makes the toolbar to be drawn when presentation is executed. Doesn't happen on "None" setting.
Here's a screenshot of what looks like:

image

adl:
I've been annoyed by #39 as well...

### DIFF
--- a/livereveal/main.css
+++ b/livereveal/main.css
@@ -79,6 +79,9 @@ div#notebook-container {
 .notebook_app {
 background-color: #ffffff;
 }
+.ctb_global_show div.ctb_hideshow.ctb_show {
+  display: none;
+}
 .cell li {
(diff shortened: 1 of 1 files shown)


damianavila:
Sound good... eventually we can add a config option to avoid hiding if somebody else request that (I can see some people changing thinks on the fly in the slideshow mode... but probably is not a common pattern), so, let's get merge this one now...

damianavila:
Thanks for you contribution!!

PR #87, opened 2015-03-24
Do not restrict the size of iframes otherwise we get a scrolbar

adl:
I'm working on a presentation where I embed some tikz code that is converted to SVG by tikzmagic.  The SVG is inserted using an iframe to work around [some issues with glyphs](http://python.6.x6.nabble.com/Problems-with-tikzmagic-and-svg-output-td5075252.html).  Unfortunately the `reset_reveal.css` style restricts the size of the iframe to 95% of its size, causing unwelcome scrollbars to appear only in reveal mode:

before

Removing this CSS entry fixes it.  The size of the iframe is already constrained by the notebook itself anyway.

### DIFF
--- a/livereveal/reset_reveal.css
+++ b/livereveal/reset_reveal.css
@@ -233,8 +233,7 @@ body {
 
 /* Ensure certain elements are never larger than the slide itself */
 .reveal img,
-.reveal video,
-.reveal iframe {
+.reveal video {
 	max-width: 95%;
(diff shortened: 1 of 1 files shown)


damianavila:
Sound reasonable, I will test it soon...

damianavila:
LGTM, let's merge it... Thanks for your contribution :+1:

PR #89, opened 2015-03-29
Fix size of texts in Button, Text, and Dropdown widgets.

adl:
I'm using RISE for interactive presentation, but the presentation is not about Python and I don't want to show any Python code.  Instead I use widgets for interactive input/output, and hide all the python code.

Unfortunately it seems the size of the text appearing in the widgets is hard-coded in the notebook style, and it is too small for a presentation.  There is also an issue with drop-down menus, because Reveal.js setups some left indentation for all `<ul>` elements.

Here are two screenshots taken in presentation mode, before and after the patch.

Before:
before

After:
after

### DIFF
--- a/livereveal/main.css
+++ b/livereveal/main.css
@@ -82,6 +82,14 @@ background-color: #ffffff;
 .ctb_global_show div.ctb_hideshow.ctb_show {
   display: none;
 }
+div.widget-area .form-control,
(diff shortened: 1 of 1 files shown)


damianavila:
Sound reasonable too... and LGTM, so let this get merge now... Thanks again for your contribution.

PR #90, opened 2015-03-29
Adjust docs to be more profile-aware.

### LINKED ISSUE #88: Explain how to configure theme in non-default profile
The readme explains how to configure the theme via `ConfigManager().update('livereveal',...)`.  However it might be nice to mention how this interacts with profiles.  Perhaps something like `ConfigManager(profile_dir=locate_profile(get_ipython().profile))`.

The motivation for this is that when I installed livereveal I used the `--profile` option, but was then puzzled when the theme configuration had no effect.  It took me some time to realize that ConfigManager operates on the default profile unless you tell it otherwise.

banbh:
The default constructor to ConfigManager refers to the default profile.
This adjusts the readme to use the active profile in ipython.  No further
comments on profiles were added as I felt these would be distracting.

Closes #88

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,14 +71,16 @@ it:
 
 ```python
 from IPython.html.services.config import ConfigManager
-cm = ConfigManager()
+from IPython.utils.path import locate_profile
+cm = ConfigManager(profile_dir=locate_profile(get_ipython().profile))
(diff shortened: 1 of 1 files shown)


damianavila:
I think is enough for now... I will merge it as is... thanks for you contribution!!

PR #92, opened 2015-03-31
hide the cell toolbox frame

adl:
Looking into why I have some frame displayed around my Markdown cells in presentation mode (see picture below), I realize that 67c92cc did not completely solve #39.

frames

In a notebook, markdown cells are not framed unless the cell toolbox is active.  So if we hide the cell toolbox, we should also hide the frame.

### DIFF
--- a/livereveal/main.css
+++ b/livereveal/main.css
@@ -82,6 +82,9 @@ background-color: #ffffff;
 .ctb_global_show div.ctb_hideshow.ctb_show {
   display: none;
 }
+.ctb_global_show .ctb_show ~ div.text_cell_render {
+  border-style: none;
+}
 div.widget-area .form-control,
 div.widget-area .dropdown-menu,
 div.widget-area .btn {


damianavila:
I don't see the difference... can you post another image with the difference?

Also, because you are there, I see a border in the cells if the celltoolbar is "hided" by RISE... maybe you can hide it here too...

border

adl:
Removing the border you see around the markdown cells is precisely what this patch is about.  Here is my example notebook after I apply the patch:

no-frame

I usually use Firefox. But since you didn't see any difference I've tested Chrome as well and it worked for me too.

damianavila:
> Removing the border you see around the markdown cells is precisely what this patch is about. Here is my example notebook after I apply the patch:

Jeje... we were talking about the same thing... going to test it today... if everything goes well, I will merge it.

damianavila:
Thanks @adl

Issue #86, opened 2015-03-24
Option to hide RISE buttons in presentation mode


juliohm:
How can I hide those `(X)` and `(?)` buttons in presentation mode?

nehalecky:
Along with @juliohm, I also am dying to know. Thanks!

damianavila:
Modifying this https://github.com/damianavila/RISE/blob/master/livereveal/main.css file to hide the buttons... something along this lines

```
#exit_b, 
#help_b {
  display: none;
}
```

should work.

juliohm:
Hi Damián, so there is no plan to add an option somewhere? Some may find
the buttons distractive.

damianavila:
I guess we can probably add a config option for that...

juliohm:
That would be great or rethink the size of the buttons.

damianavila:
Now in master you have the `,` (comma) key to toggle on/off the buttons. Closing as solved.

stoffprof:
Would it make sense to have the comma also toggle the notes/chalkboard icons in chalkboard mode?

damianavila:
> Would it make sense to have the comma also toggle the notes/chalkboard icons in chalkboard mode?

It makes sense, can you open a new issue for that @stoffprof? Thanks!

PR #93, opened 2015-03-31
toggle help and exit buttons upon pressing comma

adl:
This series of patches contain a few trivial fixes, then a way to toggle the help/exit buttons (addressing #86 in some way), and some cosmetic changes to the help dialog for key bindings.

I mapped the comma key to the help/exit toggling function because it's near the period and slash keys that are already used by Reveal for the black screen, and I think it is unlikely to conflict with any key from the notebook.  (Originally, I wanted to map it to the `?` key, unfortunately Reveal drops all key events that have a modifier like Shift pressed)


damianavila:
Nice improvements... :+1: and I like the way you solve the hiding of the buttons.
So merging! Thanks again @adl

PR #104, opened 2015-04-15
Make more options configurable

sebastianstigler:
In order to set the `width` of the slides (some beamer at work have only a low resolution), I made it and some other options configurabe like `theme`, `transition`and `start_slideshow_at`

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,9 @@ With these options, your slides will get the `serif` theme and the
 `zoom` transition, and slideshows will start from the selected cell (instead
 of from the beginning, which is the default).
 
+There are also options for `controls`, `progress`, `history`, `width`,
+`minScale` and `slideNumber`
+
 ## Usage with Leap Motion
 
 **Reveal.js** supports the [Leap Motion](leapmotion.com) controller.
--- a/README.md
+++ b/README.md
@@ -127,7 +127,10 @@ With these options, your slides will get the `serif` theme and the
 of from the beginning, which is the default).
 
 There are also options for `controls`, `progress`, `history`, `width`,
-`minScale` and `slideNumber`
+`height`, `minScale` and `slideNumber`.
+
+**Note**: The use of the `minScale` option (values other then `1.0`) can cause
+problems with codemirror.
 
 ## Usage with Leap Motion
 
--- a/README.md
+++ b/README.md
@@ -126,8 +126,21 @@ With these options, your slides will get the `serif` theme and the
 `zoom` transition, and slideshows will start from the selected cell (instead
 of from the beginning, which is the default).
 
-There are also options for `controls`, `progress`, `history`, `width`,
-`height`, `minScale` and `slideNumber`.
+You can use a similar piece of python code to change the `width` and
+`height` of your slides:
+
+```python
+from IPython.html.services.config import ConfigManager
+from IPython.utils.path import locate_profile
+cm = ConfigManager(profile_dir=locate_profile(get_ipython().profile))
+cm.update('livereveal', {
+              'width': 1024,
+              'height': 768,
+})
+```
+
+There are also options for `controls`, `progress`, `history`, `minScale` and
+`slideNumber`.
 
 **Note**: The use of the `minScale` option (values other then `1.0`) can cause
 problems with codemirror.
--- a/livereveal/main.js
+++ b/livereveal/main.js
@@ -20,8 +20,8 @@ var config = new configmod.ConfigWithDefaults(config_section, {
     controls: true,
     progress: true,
     history: true,
-    width: 1024,
-    height: 768,
+    width: 1140,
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -126,6 +126,9 @@ With these options, your slides will get the `serif` theme and the
 `zoom` transition, and slideshows will start from the selected cell (instead
 of from the beginning, which is the default).
 
+There are also options for `controls`, `progress`, `history`, `width`,
+`minScale` and `slideNumber`
+
```
damianavila: This is nice, can you add the height too?

livereveal/main.js
```diff
@@ -17,8 +17,14 @@ var config_section = new configmod.ConfigSection('livereveal',
                             {base_url: utils.get_body_data("baseUrl")});
 config_section.load();
 var config = new configmod.ConfigWithDefaults(config_section, {
+    controls: true,
+    progress: true,
+    history: true,
+    width: 1140,
```
damianavila: Please add height too...

livereveal/main.js
```diff
@@ -17,8 +17,14 @@ var config_section = new configmod.ConfigSection('livereveal',
...
+    minScale: 1.0,
```
damianavila: This sould be not modificable because of some codemirror issue, so I don't want to expose it... please remove it

damianavila:
Some minor comments... I will test it after the requested fixes... thanks!

damianavila:
Also, this will close https://github.com/damianavila/RISE/issues/97 :wink:

livereveal/main.js
adl: Wouldn't it be enough to document that setting it to another value will break codemirror?  This way, those who do not use codemirror can still play with this setting.

As far as I am concerned, I have no usage for codemirror in presentation mode (I hide all code cells and do all the interactivity via widgets), but I've had many trouble to adapt my presentation to the width of the beamer (that I cannot know before hand) because of scaling issues.

damianavila: OK, I did not think about people not showing code... the use case now makes sense... as far as it is well documented, at least with a note in the readme about the possibility to break codemirror, then I am ok to make it configurable...

sebastianstigler: I did just that with the commit 233a92f.

sebastianstigler:
Sorry,  I forget to post a comment to the second commit:

I have added height to the configurable options and added a warning for the use of the minScale option in the README.md and in the sourcecode.

damianavila:
>Sorry, I forget to post a comment to the second commit:

Yes I would love to have some kind of github ping when somebody commit to a PR... anyway, I will test on the weekend. If everything is OK (I presume that) I will merge it as is. Thanks!

livereveal/main.js
```diff
@@ -17,8 +17,15 @@ var config_section = new configmod.ConfigSection('livereveal',
                             {base_url: utils.get_body_data("baseUrl")});
 config_section.load();
 var config = new configmod.ConfigWithDefaults(config_section, {
+    controls: true,
+    progress: true,
+    history: true,
+    width: 1140,
+    height: 855, // 4:3 ratio
```
damianavila: I would like to make this width 1024 and height 768, the deafult ones are really weird values.

damianavila:
Since we are modifying this values, I would like to use "more standard" standards... 1024 x 768 seems a good choice as default.

sebastianstigler:
Changed the default values for width and height as recomended

adl:
A few weeks ago, I tried changing 1140 to 1024 because I knew 1024 was the worst-case resolution for the beamer I would use, and ran into some scalling issue I did not understand.  And unfortunately I was too much in a hurry to investigate.  Maybe the main issue was that I started building the presentation with width=1140, and so it could not fit anymore with width=1024.  One thing I understood is that 1140 is special because it is actually the width of the notebook, so maybe it helps when moving back and forth between Reveal mode and notebook mode.  I ended-up keeping 1140 in my setup.

sebastianstigler:
How about I leave the `width=1140` and `height=855` in main.js and use the 1024x768 values as examples in the README.md?

damianavila:
I don't have too much problems with width 1140... in fact I selected that in the past because of the notebook width, as @adl suggested... let's then keep it as default with height 855... and I agree with @sebastianstigler about using 1024x768 in the README so people get the idea about the configurability of the parameter to a "standard" size...

@adl, about the scaling issue you saw... this is something weird beceause we mix things reveal calculate by itself and we also impose some of the notebook calculation of top of that... so probably the behaviour is buggy and we need more investigation in the future... in the meantime, I just want to provide people with config height and width options and the then browser zooming if something does not fit well...

sebastianstigler:
And here it is...

damianavila:
Thank you very much @sebastianstigler, merging now...

PR #118, opened 2015-06-07
Workaround to the minScale != 1.0 problem after exiting presentation …

sebastianstigler:
…mode

If you set minScale to a value != 1.0, codemirror will have problem to
position the cursor via the mouse. While this problem is negligible in
presentation mode (as long as you don't want to do some live editing),
it is fatal after you exit the presentation and the cursor still behaves
strange.

This patch solves the later problem by calling Reveal.configure in the
exit function and set the minScale to 1.0. This way codemirror works
properly in none presentation mode.

Beware that there is still a cursor placing issue if you have resized
the browser window AFTER starting the presentation. If you then go back
to the none presentation mode, some cells will be rendered with a very
big height and the cursor won't be placed at the mouse cursor tip.
You can solve this by simply resizing the browser window again.


damianavila:
Sorry for the long wait... this sounds sensible... in fact, I am not sure why should people modify the minScale... I will revisit that... but merging this one. Thanks for the contribution!!
And thanks for the resize-fixing behaviour, maybe we could use that to fix the annoying bug of very big height in cells executed in presentation mode after exiting that mode... I will investigate...

Issue #123, opened 2015-07-06
enable bookmarklet use


bollwyvl:
I've recently been doing yet more notebooks-as-slides advocacy, and love showing RISE.

Working with try.jupyter.org is a great way to give people a taste with no setup, but RISE obviously isn't available there. However, with just a bit of work, RISE could be made to work entirely off of rawgit as a bookmarklet

Here's my prototype:
```html
<a href="javascript:require(["https://cdn.rawgit.com/damianavila/RISE/master/livereveal/main.js"], function(f){f.load_ipython_extension()})">
```

Everything is super duper until the static locations for some of the other assets show up:
![2015-07-06_1055](https://cloud.githubusercontent.com/assets/45380/8525230/9eb13fd4-23cd-11e5-9463-1937505844d9.png)

I think all it would take is passing `require` as the first argument to `define` in `main.js`, as [described here](http://requirejs.org/docs/api.html#modulenotes), and `toURL` will take care of the rest. Happy to PR, if this is a thing that other people would find useful!

damianavila:
>Happy to PR, if this is a thing that other people would find useful!

Of course... the PR will be welcome :wink:

damianavila:
Done!! Thanks again @bollwyvl

PR #124, opened 2015-07-07
Bookmarkletability

bollwyvl:
Implements #123, which enables use of RISE on any running Jupyter notebook via a bookmarklet.

## Try it!
Drag this _temporary_ gh-pages link (I'll take it down once there is a plan for permanent hosting) to the browser's bookmark bar.

```javascript
javascript:require(["//bollwyvl.github.io/live_reveal/livereveal/main.js"], function(r){r();});
```

I like testing it on `try.jupyter.org`, as all of my local stuff is polluted with development. You just click it, and the button shows up, hooray! Science bless `require`! I made `setup` a pun for `load_ipython_extension`, so the returned object is callable and less API-leaky.

## Hosting
On `gh-pages`, the bookmarklet gets served up all gzipped, CDNed and lovely, with the proper headers, over HTTPS, and [other](http://nytimes.github.io/svg-crowbar/) [folks](https://github.com/zeman/perfmap) [do](https://github.com/danlec/Trello-Bookmarklet) [it](https://github.com/dougal/github-bookmarklet) so I guess it's okay.

## Next steps?
I do wonder if we shouldn't load all the dependencies up front, rather than waiting for the first time the live mode is activated... it's a little weird on my slow connection, as the notebook stays in normal notebook mode and then switches over to the first slide.

Perhaps the best thing to do before documenting/publicizing would be to make and maintain a `gh-pages` (maybe [with Travis](https://gist.github.com/domenic/ec8b0fc8ab45f39403dd)) to host the bookmarklet on an `index.html`, which would include a nice draggable link in a TRY IT NOW call to action.... perhaps in a nbconverted reveal presentation with the contents of the `README`, generated from a `README.ipynb`?

Additionally, if `setup` accepted a config object, that might be a nice way to make more of the options configurable (bookmarklet generator?), even if they wouldn't live on in metadata (yet) on nbviewer/nbconvert.

### DIFF
--- a/livereveal/main.js
+++ b/livereveal/main.js
@@ -8,7 +8,8 @@
 * -----------------------------------------------------------------------------
 */
 
-define(['require',
+define([
+        'require',
         'jquery',
         'base/js/utils',
         'services/config',
@@ -155,13 +156,13 @@ function Revealer() {
   $('div#notebook-container').addClass("slides");
 
   // Header
-  $('head').prepend('<link rel="stylesheet" href=' + require.toUrl("./nbextensions/livereveal/reveal.js/css/theme/simple.css") + ' id="theme" />');
-  $('head').prepend('<link rel="stylesheet" href=' + require.toUrl("./nbextensions/livereveal/reset_reveal.css") + ' id="revealcss" />');
-  $('head').append('<link rel="stylesheet" href=' + require.toUrl("./nbextensions/livereveal/main.css") + ' id="maincss" />');
+  $('head').prepend('<link rel="stylesheet" href=' + require.toUrl("./reveal.js/css/theme/simple.css") + ' id="theme" />');
+  $('head').prepend('<link rel="stylesheet" href=' + require.toUrl("./reset_reveal.css") + ' id="revealcss" />');
+  $('head').append('<link rel="stylesheet" href=' + require.toUrl("./main.css") + ' id="maincss" />');
 
   // Tailer
-  require(['nbextensions/livereveal/reveal.js/lib/js/head.min',
-           'nbextensions/livereveal/reveal.js/js/reveal'],function(){
+  require(['./reveal.js/lib/js/head.min.js',
+           './reveal.js/js/reveal.js'],function(){
     // Full list of configuration options available here: https://github.com/hakimel/reveal.js#configuration
 
     var options = {
@@ -207,14 +208,14 @@ function Revealer() {
     dependencies: [
             //{ src: "static/custom/livereveal/reveal.js/lib/js/classList.js", condition: function() { return !document.body.classList; } },
             //{ src: "static/custom/livereveal/reveal.js/plugin/highlight/highlight.js", async: true, callback: function() { hljs.initHighlightingOnLoad(); } },
-            { src: require.toUrl("./nbextensions/livereveal/reveal.js/plugin/notes/notes.js"), async: true, condition: function() { return !!document.body.classList; } }
+            { src: require.toUrl("./reveal.js/plugin/notes/notes.js"), async: true, condition: function() { return !!document.body.classList; } }
         ]
     };
 
     // Set up the Leap Motion integration if configured
     var leap = config.get_sync('leap_motion');
     if (leap !== undefined) {
-        options.dependencies.push({ src: require.toUrl('./nbextensions/livereveal/reveal.js/plugin/leap/leap.js'), async: true });
+        options.dependencies.push({ src: require.toUrl('./reveal.js/plugin/leap/leap.js'), async: true });
         options.leap = leap;
     }
 
--- a/livereveal/main.js
+++ b/livereveal/main.js
@@ -8,10 +8,11 @@
 * -----------------------------------------------------------------------------
 */
 
-define(['jquery',
+define(['require',
+        'jquery',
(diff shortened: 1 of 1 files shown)


bollwyvl:
Quick note about `try.jupyter.org`: you have to open a given notebook in a new tab, or the javascript gets executed in the scope of the parent of the iframe, actually at `tmpnb`. We _could_ handle that particular case, but perhaps documentation is enough.

damianavila:
This is really exciting @bollwyvl!!

>On gh-pages, the bookmarklet gets served up all gzipped, CDNed and lovely, with the proper headers, over HTTPS, and other folks do it so I guess it's okay.

I know there is some service call rawgit which server github content with proper headers... I have tested it and seems to work:

```
javascript:require(["//cdn.rawgit.com/bollwyvl/live_reveal/bookmarklet/livereveal/main.js"], function(r){r();});
```

what do you think about this?

>I do wonder if we shouldn't load all the dependencies up front, rather than waiting for the first time the live mode is activated... it's a little weird on my slow connection, as the notebook stays in normal notebook mode and then switches over to the first slide.

Sound interesting... but not sure...

>Perhaps the best thing to do before documenting/publicizing would be to make and maintain a gh-pages (maybe with Travis) to host the bookmarklet on an index.html, which would include a nice draggable link in a TRY IT NOW call to action.... perhaps in a nbconverted reveal presentation with the contents of the README, generated from a README.ipynb?

Yep, it could be an option... I would love to have a way to create/load a specific notebook at `try.jupyter.org`... but I guess it is a matter of time...

>Additionally, if setup accepted a config object, that might be a nice way to make more of the options configurable (bookmarklet generator?), even if they wouldn't live on in metadata (yet) on nbviewer/nbconvert.

That's a really interesting concept I would like to see it explored!

OK, I will review in a little bit, I just wanted to give you my early feedback about this... thanks again for submit the PR!!

bollwyvl:
> I know there is some service call rawgit which server github content with proper headers... I have tested it and seems to work:

Yeah, rawgit is good, but their cdn is one-and-done: now that link will be what it was when you requested it forever. GH doesn't seem to mind hotlinking for bookmarklets, and having it on gh-pages means it's pretty easy to control when you do a new build.

> Yep, it could be an option... I would love to have a way to create/load a specific notebook at try.jupyter.org... but I guess it is a matter of time...

Again, client-side scripting might be the way: thebe and share-cell do some related stuff...

> OK, I will review in a little bit, I just wanted to give you my early feedback about this... thanks again for submit the PR!! 

No problem!

damianavila:
I tried this locally and it is not working, but with the following diff, works OK for me:
```
damian-S400CA :: /media/data/devel/jupyter_project/live_reveal/livereveal :: (pr/124)
$ git show
commit cc10ac4c13e14543e128b658d823e05a553cab53
Author: damianavila <damianavila@gmail.com>
Date:   Wed Jul 8 15:56:33 2015 -0500

    Fixing local load with require.toUrl for load reveal.js and head.min.js.

diff --git a/livereveal/main.js b/livereveal/main.js
index f2b874d..e42334b 100644
--- a/livereveal/main.js
+++ b/livereveal/main.js
@@ -161,8 +161,8 @@ function Revealer() {
   $('head').append('<link rel="stylesheet" href=' + require.toUrl("./main.css") + ' id="maincss" />');
 
   // Tailer
-  require(['./reveal.js/lib/js/head.min.js',
-           './reveal.js/js/reveal.js'],function(){
+  require([require.toUrl('./reveal.js/lib/js/head.min.js'),
+           require.toUrl('./reveal.js/js/reveal.js')], function(){
     // Full list of configuration options available here: https://github.com/hakimel/reveal.js#configuration
 
     var options = {
```

It is a simple fix, so if you have time, you can change it today... but otherwise I will merge this one and make the change by myself...

Also I would like to show this in a lightning talk at SciPy... do you agree on that?

bollwyvl:
Made the above change: I was out of pocket all afternoon.

Lightning talk away! My bad for not being there in person. Humorously, I arrived at this bookmarklet approach while working on a lightning talk url service powered by gist and nbviewer: among my targets is making authoring, administering and publishing all bookmarkletable against tmpnb. Hopefully I'll launch pretty soon!

damianavila:
@bollwyvl this is not working on `try.jupyter` for me right now... can you check? Because the last change seems OK to me...

bollwyvl:
checking it...

bollwyvl:
It won't work on `try.jupyter.org` (because of the iframe), but does appear to work on `tmpnb.org` directly.

What browser troubles you? In chrome (win/nix), the unescaped version just works. In win FF, I had to [escape it](http://www.moxleystratton.com/blog/2006/01/11/bookmarklet-compiler/):
```javascript
javascript:(function(){require%28%5B%22//bollwyvl.github.io/live_reveal/livereveal/main.js%22%5D%2C%20function%28r%29%7Br%28%29%3B%7D%29%3B}());
```

damianavila:
OK, the escape thing, we need to add it later to the README... I guess soon I have to write real docs for this :wink: merging now...

bollwyvl:
great!

the README will always strip the embedded javascript link, i think, so the
gh-pages approach is the only way you're going to get something on a github
domain with the actual, draggable bookmarklet.

PR #125, opened 2015-07-07
Update to make it work with the Jupyter notebook.

damianavila:
Some imports update and get rid of the profile concept and use a JUPYTER_CONFIG_DIR intead...

This should be merged after releasing the 3.x version, which should happen ASAP...


damianavila:
Time to merge this one...

PR #128, opened 2015-07-12
Mode transitions, LESS

bollwyvl:
Jumping between slide and notebook modes is a little jarring.

This adds some transitions, maybe too many, to the transition between the modes.

> screenshot

To achieve this, it needs a few things:
- the styles have to be loaded when modes change: this means the CSS has to be added prior to switching modes, and has to be there when you leave...
  - thus, the styles get added at extension setup time...
- to keep the styles from firing all the time against important elements like `#site`, the `body` is given a `rise-enabled` class, and all of the styles are conditional off being inside that
  - to do this, I have ported the CSS to LESS rather than putting `.rise-enabled` in front of every directive
- this also shortens the code some, as CSS not directly manipulated by reveal can be done in CSS

## tool creep
I hate tool creep as much as the next guy, but i think it makes sense in this case. LESS is a good choice because of the upstream (though with phosphor, i guess we'll have to start contending with stylus, too?). I've used `watch` to let us have live compiling without gulp/grunt.

Once the die is cast, however, things like bower may follow. even if the bower components still get checked in, they could be checked in on a `dist`-like folder so that we're not tracking all _n_000 reveal files.

### DIFF
--- a/livereveal/main.js
+++ b/livereveal/main.js
@@ -148,17 +148,17 @@ function setStartingSlide(selected) {
 
 
 function Revealer() {
+  $('body').addClass("rise-enabled");
   // Prepare the DOM to start the slideshow
-  $('div#header').hide();
-  $('div#site').css("height", "100%");
-  $('div#ipython-main-app').css("position", "static");
+  //$('div#header').hide();
+  //$('div#site').css("height", "100%");
+  //$('div#ipython-main-app').css("position", "static");
   $('div#notebook').addClass("reveal");
   $('div#notebook-container').addClass("slides");
 
   // Header
   $('head').prepend('<link rel="stylesheet" href=' + require.toUrl("./reveal.js/css/theme/simple.css") + ' id="theme" />');
   $('head').prepend('<link rel="stylesheet" href=' + require.toUrl("./reset_reveal.css") + ' id="revealcss" />');
-  $('head').append('<link rel="stylesheet" href=' + require.toUrl("./main.css") + ' id="maincss" />');
 
   // Tailer
   require(['./reveal.js/lib/js/head.min.js',
@@ -337,11 +337,9 @@ function buttonExit() {
 function Remover() {
   Reveal.configure({minScale: 1.0});
   Reveal.removeEventListeners();
+  $('body').removeClass("rise-enabled");
   $('div#site').css("height", "");
-  $('div#site').css('background-color','');
-  $("div#ipython-main-app").css("position", "");
-  $('div#header').show();
-  $('div#maintoolbar').show();
+  //$('div#maintoolbar').show();
   IPython.menubar._size_header();
 
   $('div#notebook').removeClass("reveal");
@@ -350,7 +348,6 @@ function Remover() {
   $('div#notebook-container').css('height','');
   $('div#notebook-container').css('zoom','');
 
-  $('#maincss').remove();
   $('#theme').remove();
   $('#revealcss').remove();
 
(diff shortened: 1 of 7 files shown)


damianavila:
This is really nice @bollwyvl 
Before, reviewing and merge this, I will tag a new "version": 3.x, so it get "somewhat freezed" and we can iterate on all these interesting changes.

I will give you a more detailed feedback on this PR soon. Thanks again for you work on this...

livereveal/main.js
```diff
@@ -367,7 +362,8 @@ function Remover() {
   }
 
   $('div#notebook-container').children('section').remove();
-  $('.end_space').appendTo('div#notebook-container');
+  $('.end_space').appendTo('div#notebook');
+  IPython.page.show_site();
```
damianavila: nice...

package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "fulminologist",
```
damianavila: nice name :stuck_out_tongue:

package.json
```diff
@@ -0,0 +1,18 @@
...
+  "version": "0.0.0",
+  "description": "",
```
damianavila: fill description...

package.json
```diff
@@ -0,0 +1,18 @@
...
+  "main": "index.js",
```
damianavila: `index.js`?

package.json
```diff
@@ -0,0 +1,18 @@
...
+  "scripts": {
+    "less": "PATH=./node_modules/.bin:$PATH lessc --autoprefix src/less/main.less livereveal/main.css",
```
damianavila: This need docs at the readme (well, we have to build a doc, but for the moment README is enough...

package.json
```diff
@@ -0,0 +1,18 @@
...
+    "watch-less": "./node_modules/.bin/watch 'npm run less' src/less"
+  },
+  "author": "",
```
damianavila: you!!

src/less/main.less
```diff
@@ -0,0 +1,190 @@
+@some-grey: #727272;
+@some-other-grey: #dfdfdf;
+
+.transition-all {
```
damianavila: you put `.transition-all` in a lot of places, can you explain me why? (I have a lot to learn about css yet!!)

src/less/main.less
```diff
@@ -0,0 +1,190 @@
...
+
+  #notebook-container {
+    .transition-all;
+    background-color: transparent;
+    box-shadow: none;
+    width: 100%;
+  }
+
+  .ctb_global_show {
+    .ctb_hideshow.ctb_show { display: none; }
```
damianavila: this seems to be not working, because if I enter the slideshow mode with the ctb active, I still see it...

damianavila:
This looks very nice, I left some comments...

package.json
bollwyvl: Opps!

package.json
bollwyvl: Just the default from `bower init`... will correct with other stuff.

src/less/main.less
bollwyvl: This is a mixin for making all properties transtionable. we can turn off particular ones if wanted.

docs:
https://developer.mozilla.org/en-US/docs/Web/CSS/transition-property

damianavila: Thanks!!

damianavila:
OK, I think this is good to go... let's merge it! Thanks you!

Now we need to continue with:
>Once the die is cast, however, things like bower may follow. even if the bower components still get checked in, they could be checked in on a dist-like folder so that we're not tracking all _n_000 reveal files.

:stuck_out_tongue_winking_eye:

PR #132, opened 2015-08-03
Don't explicitly supply config dir to enable extension

takluyver:
It was changing the config in the wrong directory - see jupyter/notebook#251 for details.

If people want to install with different config directories, they should set the JUPYTER_CONFIG_DIR environment variable. ConfigManager will respect this automatically, without you needing to do anything.

i.e. instead of `python setup.py install --jupyter-config-dir foo/`, get people to do:

    JUPYTER_CONFIG_DIR=foo/ python setup.py install

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,10 @@ To install this nbextension, simply run ``python setup.py install`` from the
 *RISE* repository.
 
 The setup script will respect your `JUPYTER_CONFIG_DIR` environment variable to
-enable the extension.
+enable the extension. So to enable the extension for a non-default config
+directory, run:
+
+    JUPYTER_CONFIG_DIR=path/to/config/ python setup.py install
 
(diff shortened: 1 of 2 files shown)


damianavila:
LGTM, can you add the info to the README? (I have updated just now with other things, so you probably need to pull again before making any changes :wink:)

takluyver:
Done.

damianavila:
Can I be a little annoying and ask you to add `JUPYTER_CONFIG_DIR=foo/ python setup.py install` to the README? Some people using the extension maybe does not know how setup.py "respect" an env variable... I know for other could be trivial, but I also know some people will not understand this with an explicit command... Thanks!

takluyver:
You can - there you go :-)

damianavila:
Thanks!

PR #156, opened 2015-10-11
Alternative configuration method using notebook metadata

fcremo:


livereveal/main.js
```diff
@@ -32,7 +30,15 @@ var config = new configmod.ConfigWithDefaults(config_section, {
     slideNumber: true,
     start_slideshow_at: 'beginning',
     scroll: false,
-});
+};
+
+if(Jupyter.notebook.metadata.livereveal != undefined){
```
damianavila: I prefer to keep `IPython` for now (to keep consistency) and use `!==`

```javascript
if (IPython.notebook.metadata.livereveal !== undefined) {
```

livereveal/main.js
```diff
@@ -32,7 +30,15 @@ var config = new configmod.ConfigWithDefaults(config_section, {
...
+    default_config = $.extend(true, default_config, Jupyter.notebook.metadata.livereveal);
+} else {
+    config_section.load();
```
damianavila: This should be before the `if`, otherwise, if you trigger the condition, it is never executed... why you put it here?

fcremo:
You're right, there's no reason to put it there, it was some ugly code that I wrote while experimenting. Fixed everything in https://github.com/fcremo/RISE/commit/a1152ccebd4f2ad66c17cb4153e2c551038ea165
Do you need me to make another pull request?

damianavila:
LGTM, merging now...

PR #164, opened 2015-11-06
Conda recipe for RISE

ijstokes:
Just tested on my OS X 10.10 laptop using latest Jupyter conda package as base.  But seems to work.

### DIFF
--- a/conda.recipe/build.sh
+++ b/conda.recipe/build.sh
@@ -1,0 +1,3 @@
+SRC_DIR=$RECIPE_DIR/..
+cd $SRC_DIR
+python setup.py install
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -1,0 +1,23 @@
+package:
+  name: rise
+  version: 0.3.0
+
+build:
+    number: 1
+
+source:
+  git_url: https://github.com/ijstokes/RISE.git
+  git_tag: master
+
+requirements:
+  build:
+    - python
+    - jupyter
+    - notebook
+  run:
+    - python
+    - jupyter
+    - notebook
+
+about:
+  home: https://github.com/damianavila/RISE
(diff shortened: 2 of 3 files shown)


ijstokes:
The `meta.yaml` file specifies my forked repo GH URL, instead of the actual project repo.

conda.recipe/meta.yaml
```diff
@@ -0,0 +1,23 @@
...
+    number: 1
+
+source:
+  git_url: https://github.com/ijstokes/RISE.git
+  git_tag: master
+
+requirements:
+  build:
+    - python
+    - jupyter
```
damianavila: jupyter is the metackage installing all... there is no need to add it here... only the notebook should be necessary

conda.recipe/meta.yaml
```diff
@@ -0,0 +1,23 @@
+package:
+  name: rise
+  version: 0.3.0
+
+build:
+    number: 1
+
+source:
+  git_url: https://github.com/ijstokes/RISE.git
```
damianavila: This has to point to the main repo... as you mnetioned in some comment...

conda.recipe/meta.yaml
```diff
@@ -0,0 +1,23 @@
+package:
+  name: rise
+  version: 0.3.0
```
damianavila: This should be `3.x`... in fact, we probably need to add a `version` file somewhere in the future, but that's for later...

conda.recipe/build.sh
```diff
@@ -0,0 +1,3 @@
+SRC_DIR=$RECIPE_DIR/..
+cd $SRC_DIR
```
damianavila: you are missing npm-related steps here... check: https://github.com/damianavila/RISE#development (the same goes for the `bld.bat` file.

damianavila:
@ijstokes I left some comments here...

damianavila:
@ijstokes closing in favor of #204
Thanks for your contribution!

PR #176, opened 2015-12-23
Making installation as user non-default

maartenbreddels:
This makes it more like a standard way of installing:
$ python setup.py install
and for a user:
$ python setup.py install --user


damianavila:
The new packaging/installing mechanism in #203 should handle this --user case, so closing here.
Thanks for your contribution @maartenbreddels

maartenbreddels:
:+1:

PR #200, opened 2016-07-09
Added companion CSS file

jason-s:
I would like to be able to place a .css file alongside the .ipynb file to override the theme in a few ways. This allows private modifications without having to find or change the installed reveal.js themes.

### DIFF
--- a/livereveal/main.js
+++ b/livereveal/main.js
@@ -174,6 +174,12 @@ function Revealer() {
   // Header
   $('head').prepend('<link rel="stylesheet" href=' + require.toUrl("./reveal.js/css/theme/simple.css") + ' id="theme" />');
   $('head').prepend('<link rel="stylesheet" href=' + require.toUrl("./reset_reveal.css") + ' id="revealcss" />');
+  if (window.location.pathname.endsWith('.ipynb'))
+  {
+    // Attempt to load CSS with the same path as the .ipynb but with .css extension instead
+	var cssfile = window.location.pathname.replace(/\.ipynb$/,'.css');
+	$('head').append('<link rel="stylesheet" href=' + cssfile + ' id="notebook-custom-css" />');
+  }
 
   // Tailer
   require(['./reveal.js/lib/js/head.min.js',


jason-s:
Question: should the code look like this (which is what I've committed)

```
  // Header
  $('head').prepend('<link rel="stylesheet" href=' + require.toUrl("./reveal.js/css/theme/simple.css") + ' id="theme" />');
  $('head').prepend('<link rel="stylesheet" href=' + require.toUrl("./reset_reveal.css") + ' id="revealcss" />');
  if (window.location.pathname.endsWith('.ipynb'))
  {
    // Attempt to load CSS with the same path as the .ipynb but with .css extension instead
	var cssfile = window.location.pathname.replace(/\.ipynb$/,'.css');
	$('head').append('<link rel="stylesheet" href=' + cssfile + ' id="notebook-custom-css" />');
  }
```

or this

```
  // Header
  if (window.location.pathname.endsWith('.ipynb'))
  {
    // Attempt to load CSS with the same path as the .ipynb but with .css extension instead
	var cssfile = window.location.pathname.replace(/\.ipynb$/,'.css');
	$('head').prepend('<link rel="stylesheet" href=' + cssfile + ' id="notebook-custom-css" />');
  }
  $('head').prepend('<link rel="stylesheet" href=' + require.toUrl("./reveal.js/css/theme/simple.css") + ' id="theme" />');
  $('head').prepend('<link rel="stylesheet" href=' + require.toUrl("./reset_reveal.css") + ' id="revealcss" />');
```

The 2nd option puts the custom css higher in the list which means it doesn't override later CSS files, only the reveal.js theme.

damianavila:
>Question: should the code look like this (which is what I've committed)

I have to think about this... and also in the general approach to inject custom.css... but first, thanks for the PR!

damianavila:
I was thinking about this and I think we need a more general solution for this... do you want different customization per notebook? In this case, something like this would help... BUT, if you want to style a set of slides with the same css, in that case a rise.css into the jupyter share directory would be way to do it, so it get's applied to all the slides used with he RISE installation living in a specific environment (or the main/root installation if you are not using environments at all). What use case do you want to support?

jason-s:
Both, although I would generally use the latter:

- the "global" approach ("a rise.css into the jupyter share directory would be way to do it") is good for invididuals
- the "sidecar" approach is better for teams or cases where the .css file belongs in version control along with the notebook file. Even if this duplicates the css files, it means the two files provide sufficient information on their own; you don't have to tell someone to install rise.css file from website ABC, since it might conflict with another rise.css file.

damianavila:
>Both, although I would generally use the latter:

OK, I see scenarios for both case as you, can you fix the conflict in the PR? Thanks!

jason-s:
> can you fix the conflict in the PR? Thanks!

I can, it might be several weeks before I can set aside time to do this however, so don't treat the delay as a lack of interest, just as a lack of time.

damianavila:
No problem, I can do it by myself, thanks for let me know about this :wink:

damianavila:
Closing as superseded by @296. Thanks @jason-s!

Issue #172, opened 2015-12-03
Editing livereveal  in notebook metadata


btengels:
Trying to add the following snippit to the metadata

    "livereveal": {
        "theme": "sky",
        "transition": "none"
    },

But nothing changes. In fact, it doesn't matter what the 'theme' or 'transition' values are set to. Tested in Chrome and firefox. Pasted right before the "kernelspec" block as suggested in the readme.

btw, Damián this project is really the greatest. Bien hecho.

bs1234:
I found a similar issue with transitions and width / height as well.
In case it helps you: when i shutdown my notebook and restarted the changes were in effect.
Might or might not be similar for you. Might or might not be something to do with caching.

damianavila:
yep, caching could be an issue, try hard-reloading the page, disabing/cleaning cache or in incognito mode...

parmentelat:
same issue here, it helps to know it's a matter of caching, although some deterministic method to get the new attributes to kick in would be helpful

regardless, this is an awesome tool, thanks a million for putting this together

frederikph:
I am having a similar issue. Unfortunately, restarting the notebook and clearing cache doesn't work.

jason-s:
I ran across the problem as well. There's a bug, which I figured out when I added some `console.log` statements in `main.js`.

This section of code:

```
config_section.load();

if(IPython.notebook.metadata.livereveal !== undefined){
    default_config = $.extend(true, default_config, IPython.notebook.metadata.livereveal);
}
var config = new configmod.ConfigWithDefaults(config_section, default_config);
```

the if statement is never true. It appears as though it executes before the metadata is actually loaded from the notebook.

Same with the `setup()` function, that runs before the notebook metadata is loaded.

Not sure when the right time is to grab the metadata and use them. See https://github.com/ipython/ipython/issues/7528

damianavila:
It seems a bug as you said... I will look into your proposal @jason-s

damianavila:
People, I put a PR to solve this, check: https://github.com/damianavila/RISE/pull/201

BTW, @jason-s, I took another path to solve it but thanks for your interest and discussion!!

Issue #194, opened 2016-06-15
Changing configuration doesn't affect unless re-opening the notebook


jluttine:
I have the following in one of my cells:
```
from notebook.services.config import ConfigManager
cm = ConfigManager()
cm.update('livereveal', {
              'width': 1024,
              'height': 768,
})
```
I execute the cell and enter RISE mode, but it has no effect. If I close the tab and re-open the notebook, then it takes effect. I think executing config cells should affect immediately?

damianavila:
Since the params at read from the config when the notebook load, you need to reload the page to apply the changes... but we should be loading the config before entering the slideshow mode, so we can change things immediately...

damianavila:
I put a PR to solve this, check: https://github.com/damianavila/RISE/pull/201

damianavila:
PR merged, closing here.

PR #201, opened 2016-07-16
Fix custom config things

damianavila:
Also making config possible without restarting the server.

This should close: #172 and #194


damianavila:
Let's start merging things...

Issue #169, opened 2015-11-13
versioning...


damianavila:
We need something to track dev versions and official releases...

damianavila:
PR merged, closing this one as well...

Issue #147, opened 2015-09-28
Installing Rise with pip?


tschm:
Hi, I try to download and install Rise using pip (I would strongly prefer anaconda). Here's what I get...

Collecting git+http://github.com/damianavila/RISE.git (from -r /home/thomas/github/pymosek/requirements.txt (line 2))
  Cloning http://github.com/damianavila/RISE.git to /tmp/pip-KOo_6K-build
    Complete output from command python setup.py egg_info:
    usage: -c [-h] {install} ...
    -c: error: invalid choice: 'egg_info' (choose from 'install')
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 2 in /tmp/pip-KOo_6K-build

damianavila:
@tschm Currently working in a branch to get a pip installable thing and conda packages as well...
I will let you know when it is ready...

DavidPowell:
I am also keen to get a one line installation solution based on pip or conda, because I am trying to include RISE into a docker image I am creating, which is currently a bit complex.

damianavila:
I know... I will be in vacation for a week, but I plan to deal with this when I come back...

TMorville:
Hi Damian, and thanks for the effort! 

I am on the same wagon. After switching from Canopy to Anaconda (to get Python 3.x) I am getting used to installing with conda. 

If i use python setup.py install RISE does not seem to get registred in the Anaconda environment, even though my "which python" points to the anaconda bin. 

Anyway, a conda compatible installation would be great! Looking forward to using your package.

akaihola:
@damianavila, could you push that branch here? I'd love to contribute if needed.

ThomasA:
+1

akaihola:
@damianavila, also, do you think this could be used: https://github.com/jdfreder/jupyter-pip

akaihola:
I created an experimental pip compliant `setup.py`, see https://github.com/akaihola/RISE/commits/master

damianavila:
PR merged, closing here...

Issue #52, opened 2015-02-01
We need to support a multiplatform installation


damianavila:
Some things to discuss/implement:

* [x] remove dual installation... just let the python setup.py
* [X] also support win
* [x] support pip install
* [x] conda package
* [x] local and system wide installation
* ~~how to deal with require in custom.js (I like the current solution although is hacky)~~

Other ideas welcome...

damianavila:
* [ ] version number has to be standard, 3.x will be problematic, let's make it real and maybe turn into 0.3 and redo the previos releases as 0.1 and 0.2 (yes I know... it was a mistake trying to use 1.x, 2.x, etc)

bollwyvl:
Sheepishly, I must admit I did a bunch of [cowboy stuff](https://github.com/bollwyvl/live_reveal/tree/bridge-too-far) without doing incremental commits and changing too many names: it's kind of broken, but parts of it might be salveagable.

I kind of like what @takluyver did with [cite2c](https://github.com/takluyver/cite2c/blob/master/install.py). Frankly, I didn't even know `load_extensions` was a thing: seems pretty clean. At any rate, it should be one-liner `require`-able, even if config of certain things was still possible.

To reduce the focus on fiddling with configuration files, I think any slideshow customizations (`transition`, etc.) could move into notebook metadata... here's the thing I've been hacking on:
![](http://i.imgur.com/eLXugBX.png)
While I started building it [against IPython](https://github.com/bollwyvl/ipython/compare/slideshow-moremeta) to learn about actions, perhaps `live_reveal` would be a more fulfilling place to put it, as a lot of the options don't make any sense unless you can actually see it working... of course, `nbconvert` would have to know about it for it to be worth anything... but that's a good thing to shoot for!

Version: what's wrong with `3.x`? As IP 3 is imminent and reveal 3 is already out, it seems kind of appropriate.

Interested to see where this all goes!

takluyver:
I have, for now, completely bypassed packaging systems in the cite2c install script, though that may have to change to handle the server side extension I've added recently. The load_extensions thing is nice because the install script can enable the new extension, but you can't do that as part of conda packaging or pip installation from wheels, because those can only place new files on the filesystem. There may be less need for this if @juhasch's [config extension](https://github.com/ipython-contrib/IPython-notebook-extensions/wiki/config-extension).

[...]

PR #203, opened 2016-07-25
Improve setup.py and update installation mechanism

### LINKED ISSUE #56: Make sure we use all the new potential behind the nbextension system
A lot of things changed in the last months... we have to update the code to make it more compatible with the current possibilities of the nbextensions architecture...

damianavila:
This should enable a pip installable package compatible with the 4.2 notebook mechanism to deal with nbextensions. Additionally, some renaming was done.
Next step should be test that it actually work installing from pypi and add a conda package (maybe in another PR)

This one should close #56 #147 #169 and make advances in #52

And this is on top of #201


damianavila:
Let's merge this one as well...

PR #205, opened 2016-08-01
Closes #188: calling Reveal.sync after output is generated

### LINKED ISSUE #188: Interactive code cell output effects on vertical centering/arrangement of slide content
Hi,

RISE is fantastic and allows the creation of interactive presentations, which is really great!

So, the interactive part is, for me, something really important and I am having troubles with it. This concerns the slide organization and code cell output. In a case where the output hasn't been generated, the slide is organized considering the current status, and outputs generated on-the-fly/in presentation mode may push contents outside the slide. This has to do with the vertical centering of the slide **not** being refreshed after the output is generated (I am talking about cases where the cells with output included should fit the slide). I don't know if this refresh is possible.

In alternative, if it not possible, I think that toggled output should work as a way to circumvent this. The slide mode should arrange the cells on the slide taking the content of toggled outputs into consideration. This would allow slides that are ready for the interactive generation of output, while not messing with the structure of the slide. Currently, toggled output gives the same slides as ungenerated output.

I hope my explanation is clear.

Best,
João

fjsj:
Closes #188  
This makes Reveal.js to vertically center slides considering outputs generated on the fly.

### DIFF
--- a/rise/static/main.css
+++ b/rise/static/main.css
@@ -221,7 +221,7 @@
   font-size: 11px;
   background-color: #f7f7f7;
   color: #333;
-  box-shadow: 0 1px 0 rgba(0, 0, 0, 0.2), 0 0 0 2px #ffffff inset;
+  box-shadow: 0 1px 0 rgba(0, 0, 0, 0.2), 0 0 0 2px #fff inset;
   border-radius: 3px;
   display: inline-block;
   margin: 0 .1em;
--- a/rise/static/main.js
+++ b/rise/static/main.js
@@ -11,10 +11,9 @@
 define([
         'require',
         'jquery',
-        'underscore',
         'base/js/utils',
         'services/config',
-], function(require, $, _, utils, configmod) {
+], function(require, $, utils, configmod) {
 
 function configSlides() {
   /*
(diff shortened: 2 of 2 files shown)


livereveal/main.css
```diff
@@ -221,7 +221,7 @@
   font-size: 11px;
   background-color: #f7f7f7;
   color: #333;
-  box-shadow: 0 1px 0 rgba(0, 0, 0, 0.2), 0 0 0 2px #ffffff inset;
+  box-shadow: 0 1px 0 rgba(0, 0, 0, 0.2), 0 0 0 2px #fff inset;
```
damianavila: if there is something to change at the css level, you should modify src/main.less because this one is generated by that file...

damianavila:
Thanks @fjsj, I will take a look on this!!

livereveal/main.css
fjsj: I changed nothing in less/css. This was a change that my less compiler introduced when I ran `npm run build`. Mine is lessc 2.5.3 on Mac OS.

fjsj: Also it's strange because in other places of the same file it's #fff and here it's #ffffff. Maybe it was a lessc < 2.5.3 bug? If you want I can revert this.

damianavila: that's ok... if it was automatically generated, I am ok with the change...

damianavila:
@fjsj I have tested this PR and I found that the slide content is relocated to "center" the whole thing in the case that **originally** the output fit on the slide. If **originally** the content does not fit into the slide, the fix does not help. I think your proposal make things better, so I am willing to merge it, but I want to check that is the intended behavior and I am not missing something in my testing...

fjsj:
@damianavila this PR should fix the case where there is space above the content. Without this fix, even if there is space above the content, the output would be rendered below and outside the visible area. With this fix, the whole content is relocated to "center", so if there is space above it will be filled.

damianavila:
> The idea here is for the presenter to organize the slides with all outputs generated, but clearing them before presentation. That way the presenter would be able generate outputs on-the-fly without worrying about them being partially visible.

OK, got it... happy to merge this one when you fix the merge conflicts :wink:

fjsj:
@damianavila done rebasing

damianavila:
OK, LGTM, merging this one now. Thanks for your contribution!

fjsj:
You're welcome, thanks too!

j3mdamas:
Thank you! I'll test it when I have some time.

PR #218, opened 2016-08-19
Add support for headers and footers on each slide

anhi:
Hi,

I needed a way to add consistent headers and footers to my slides. The approach is not particularly elegant, but works well enough for me. I've just added four configuration options to the metadata where I can specify html to be inserted by Revealer() and removed by Remover():

````
  "livereveal": {
    "headerLeft": "`header left",
    "headerRight": "header right",
    "footerLeft": "footer left",
    "footerRight": "footer right"
  },
```

I'm not sure if I chose the correct div to append the headers and footers to, but they end up where I want them...

### DIFF
--- a/rise/static/main.js
+++ b/rise/static/main.js
@@ -32,6 +32,10 @@ function configSlides() {
       slideNumber: true,
       start_slideshow_at: 'beginning',
       scroll: false,
+      headerLeft: '',
+      headerRight: '',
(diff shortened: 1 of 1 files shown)


damianavila:
Thanks! Probably not elegant but interesting... I will take a look!

anhi:
Superseded by #222

PR #226, opened 2016-09-09
Added support for a chalkboard plugin

superbobry:
The current limitations are listed in the JS source.


rise/static/main.js
```diff
@@ -387,41 +403,31 @@ function buttonHelp() {
     var help_button = $('<i/>')
         .attr('id','help_b')
         .attr('title','Reveal Shortcuts Help')
-        .addClass('fa-question fa-4x fa')
+        .addClass('fa-question fa-2x fa')
         .addClass('my-main-tool-bar')
-        .css('position','fixed')
```
superbobry: I've moved these properties to the CSS. LMK if I should revert this and the following hunks back.

rise/static/main.js
```diff
@@ -257,25 +255,39 @@ function Revealer(config) {
...
     76: null, // l, right disabled
-    78: null, // n, down disable
     79: null, // o disabled
     80: null, // p, up disable
     // 83: null, // s, notes, but not working because notes is a plugin
     87: function() {Reveal.toggleOverview();}, // w, toggle overview
     188: function() {$('#help_b,#exit_b').fadeToggle();},
     },
 
+    // Current issues:
```
superbobry: ^^^

damianavila:
Great!! I will take a look as soon as possible (that could be some days...)

superbobry:
Two more issues to be aware of

* The plugin doesn't work for fragment->fragment transitions, see rajgoel/reveal.js-plugins#14.
* It does not yet provide a way to customize line width, although this is trivial to implement.

damianavila:
Closing this one in favor of https://github.com/damianavila/RISE/pull/355.
Thanks @superbobry for bringing this idea and implementation.

PR #227, opened 2016-09-10
Hide Input/Output number when presenting

superbobry:
I don't think they actually add anything to the content. And by the way `nb-present` does that too.

### DIFF
--- a/rise/static/main.css
+++ b/rise/static/main.css
@@ -165,13 +165,11 @@
 .rise-enabled .text_cell {
   font-size: 160%;
 }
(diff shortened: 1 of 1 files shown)


damianavila:
This should be a configurable option. And probably needs to live in the rise.js to make it configurable.

parmentelat:
+1 for making this configurable, personally I would like to keep the In: Out: business if that's what we're talking about; it's helpful if only to show that the interpreter is busy looping..

damianavila:
Let's close this one as is and implement it as a configurable thing. Opening #386 to track it.

Issue #166, opened 2015-11-09
Vertical centering seems impossible to disable from Rise


pmjandreen:
Reveal has the following option for choosing vertical alignment style:

```
// Vertical centering of slides
center: true
```

I find that when I try to set center to false using Rise it has no effect (I tried both ways of configuring Reveal from Rise).

damianavila:
It could be some problems with the notebook css... this need to be investigated...

damianavila:
@pmjandreen how did you try to set this false... my best guess is modifying the source code. But I would like confirmation before proceeding with the discussion and possible fix/enhancement.

pmjandreen:
I tried both ways of configuring Reveal from Rise (in the code and in the metadata of the notebook).

damianavila:
@pmjandreen thanks for the info, I will take a look.

kevin-keraudren:
In order to disable centering, you need to set if to `false` in reveal.js, as you already did, but you also need to add a few lines to `~/.jupyter/custom/custom.css`:

```css
.my-outer-container {
    width:100%;
    height:100%;
    position:absolute;
    display:table;
}
.my-center-container {
    display:table-cell;
    vertical-align:middle;
    width:100%;
    height:100%:
}
```

see notebook here https://github.com/kevin-keraudren/pydata-meetup-20161101 which gets rendered like this http://www.slideshare.net/kevinkeraudren/medical-imaging-in-python-from-morphological-operations-to-machine-learning-and-visualisation

damianavila:
@kevin-keraudren, nice info, thanks!

akaihola:
@kevin-keraudren, I didn't have to modify CSS after I applied a patch to RISE (see push above your comment).

damianavila:
> I applied a patch to RISE (see push above your comment).

Can you link the patch? Thanks!

akaihola:
I think GitHub already linked it for us above, but here you go, the commit is 1bb0261f145c0dca85e2a7dcb5c932ca7ccef0d1. I made pull request #232 for it.

damianavila:
> I think GitHub already linked it for us above

Thanks for the reference... I missed somehow.

damianavila:
PR was merged and tested OK, closing here.

PR #232, opened 2016-10-23
Pass the "center" option to Reveal.configure() (#166)

akaihola:
The `center` option was simply omitted from `Reveal.configure()`, and thus had no effect in the notebook metadata or `ConfigManager.update()`.

### DIFF
--- a/rise/static/main.js
+++ b/rise/static/main.js
@@ -32,6 +32,7 @@ function configSlides() {
       slideNumber: true,
       start_slideshow_at: 'beginning',
       scroll: false,
+      center: true
   };
 
   var config_section = new configmod.ConfigSection('livereveal',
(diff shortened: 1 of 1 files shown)


damianavila:
Thanks for the PR, looking at it soon.

rise/static/main.js
```diff
@@ -32,6 +32,7 @@ function configSlides() {
       slideNumber: true,
       start_slideshow_at: 'beginning',
       scroll: false,
+      center: true
```
damianavila: I am open to add this option, but I would prefer `center: false`, otherwise we are changing the `status quo`.

damianavila:
@akaihola I just left a comment, sorry to the long time I took to give you some feedback :wink:

rise/static/main.js
damianavila: `center` is true by default, forget my previous comment.

damianavila:
LGTM, merging now.

Issue #233, opened 2016-11-02
Vertical centering, code text cut off, scrolling and z-index


kevin-keraudren:
Hi,

I recently did a presentation using RISE, thanks a lot for this great addition to the Jupyter project.

I did a few CSS tweaks you might be interested in.

I already posted comments regarding:
 * vertical centering of slides https://github.com/damianavila/RISE/issues/166
 * code text cut off on wide slides https://github.com/damianavila/RISE/issues/129

There are other minor modifications I did here:
https://github.com/damianavila/RISE/compare/master...kevin-keraudren:klarismo-chrome-changes

In particlar, here is how I addressed the vertical scrolling issue in Chrome (for Safari, setting the `scroll` option to `true` is enough):
https://github.com/kevin-keraudren/RISE/commit/5ecb59b11f47ac05a660d44df74d998e6cd18726
This issue is similar to https://github.com/jquery/jquery/issues/2215 .
There are two open tickets about it, https://github.com/damianavila/RISE/issues/229 and https://github.com/damianavila/RISE/issues/185 .

I would also suggest adding `.css('z-index', '30')` to the exit and help button so they always stay in the foreground.

The repository for my slides is available here https://github.com/kevin-keraudren/pydata-meetup-20161101 and you can see on SlideShare how they are rendered on my laptop http://www.slideshare.net/kevinkeraudren/medical-imaging-in-python-from-morphological-operations-to-machine-learning-and-visualisation

damianavila:
A lot of interesting feedback and solutions!! Thanks for trying RISE and look into the problems.
I will try to review your PRs soon.

kevin-keraudren:
There is only one trivial PR opened for now regarding the z-indices.

The proposed fix for scrolling in Chrome (namely removing `overflow: hidden;` in the CSS of the body) seems to be due to a bug in Chrome according to the comments in the jquery link, so I do not know if it is a legitimate PR to open.

damianavila:
> The proposed fix for scrolling in Chrome (namely removing overflow: hidden; in the CSS of the body) seems to be due to a bug in Chrome according to the comments in the jquery link, so I do not know if it is a legitimate PR to open.

Maybe, but we should test how this will work for Firefox or make a "selective" PR to add that piece on a "browser-dependent" way until it is fixed upstream.

damianavila:
Referenced PRs were merged, issues closed and now scrolling works reliable (at least in my tests :wink:) in the latest release. Closing here.

PR #234, opened 2016-11-02
Set help and exit buttons in foreground with z-index

kevin-keraudren:
Add z-index to CSS of help and exit buttons so that they are always in the foreground, as suggested in https://github.com/damianavila/RISE/issues/233

Before:
screen shot 2016-11-02 at 17 00 22

After:
screen shot 2016-11-02 at 16 59 33

Of course it is a question of preference, except that if you do not set the `z-index` of the buttons, they will be rendered behind the cells of code, but in front of the rendered markdown, which is not very consistent...

### DIFF
--- a/rise/static/main.js
+++ b/rise/static/main.js
@@ -393,6 +393,7 @@ function buttonHelp() {
         .css('bottom','0.5em')
         .css('left','0.6em')
         .css('opacity', '0.6')
+        .css('z-index', '30')
         .click(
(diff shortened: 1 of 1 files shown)


damianavila:
Sounds like a proper fix, testing it soon.

damianavila:
Some months later but finally tested and works OK, so merging.
Thanks @kevin-keraudren.

Issue #102, opened 2015-04-07
Let start_slideshow_at =  'selected' also exit the slideshow at the selected cell


jorisvandenbossche:
You now have the option of letting your slideshow start at the selected cell when going into slide show mode (https://github.com/damianavila/RISE#configure-your-own-options).

I wondered if it could also be an option (or combined with `start_slideshow_at`) to go to the selected cell when exiting RISE, instead of going back to the beginning of the notebook.

damianavila:
Probably easy to do, I marked as enhancement...

aaronbarzilai:
+1, I was just about to make the same request when I found this thread

parmentelat:
+1 - as originally stated in #241 

this would really be a life saver as far as I am concerned :)

damianavila:
Nicely and thoughtfully implemented by @parmentelat 
Closing here.

PR #251, opened 2017-03-31
a rough implementation of enhancement described in #102

parmentelat:
followup for #102 

before exiting reveal mode, this patch computes
the notebook index of the first cell in the currently displayed slide
that cell then gets selected and focused when returning to normal
notebook mode

this first attempt implements this behaviour unconditionally,
i.e. regardless of `start_slideshow_at`
I am not convinced that it should depend on that setting anyway
point being that it is rather straightforward to go to the beginning or end 
of the presentation when in normal notebook mode, and generally not critical 

this is of course open to further discussion though

### DIFF
--- a/rise/static/main.js
+++ b/rise/static/main.js
@@ -462,6 +462,54 @@ function Remover(config) {
   disconnectOutputObserver();
 }
 
+// just before exiting reveal mode, we run this function
+// whose job is to find the notebook index
+// for the first cell in the current (sub)slide
+// this allows to restore the notebook at the correct location,
+// i.e. with that cell being selected
+//
+// we use the current URL that ends up in 'slide-n-m'
+// to find out about the slide and subslide 
+function reveal_cell_index(notebook) {
+  // last part of the current URL holds slide and subslide numbers
+  var href = window.location.href;
+  var chunks = href.split('-');
+  var len = chunks.length;
+  var slide = Number(chunks[len-2]);
+  var subslide = Number(chunks[len-1]);
+
+  // just scan all cells until we find one at that address
+  // except that we need to start at -1 0r 0 depending on
+  // whether the first slide has a slide tag or not
+  var is_slide = function(cell) {
+    return cell.metadata.slideshow
+	&& cell.metadata.slideshow.slide_type == 'slide';
+  }
+  var is_subslide = function(cell) {
+    return cell.metadata.slideshow
+	&& cell.metadata.slideshow.slide_type == 'subslide';
+  }
+  var slide_counter = is_slide(notebook.get_cell(0)) ? -1 : 0;
+  var subslide_counter = 0;
+  var result = null;
+	
+  notebook.get_cells().forEach(function (cell, index) {
+    if (result) 
+      // keep it short: skip if we found already
+      return;
+    if (is_slide(cell)) {
+      slide_counter += 1;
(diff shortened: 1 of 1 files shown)


parmentelat:
As far as testing:
* I have only been able to try it out in my own environment, it seems to be working as I expect it to, at least
* I have not been able to spot any other tests to run, please provide instructions if relevant

parmentelat:
One last thing, as compared to what the original issue was requesting:
* I do not even look at what the 'selected' cell is in the notebook when reveal is turned off
* this is because most of the time there is no such selection
* instead I try to locate  the (first cell in the) slide that is currently displayed

damianavila:
Thanks for the implementation, I will try to take a look as soon as I can.

rise/static/main.js
```diff
@@ -483,13 +531,17 @@ function revealMode() {
     buttonHelp();
     $('#maintoolbar').addClass('reveal_tagging');
   } else {
+    var current_cell_index = reveal_cell_index(IPython.notebook);
```
damianavila: You custom function is not necessary, I think... you can use something like this:
```javascript
var current_cell_index = IPython.notebook.get_selected_index()
```

rise/static/main.js
```diff
@@ -483,13 +531,17 @@ function revealMode() {
...
     Remover(config);
     setupKeys("notebook_mode");
     $('#exit_b').remove();
     $('#help_b').remove();
     $('#maintoolbar').removeClass('reveal_tagging');
     // Workaround... should be a better solution. Need to investigate codemirror
     fixCellHeight();
+    // select and focus on current cell
+    IPython.notebook.select(current_cell_index);
+    IPython.notebook.get_selected_cell().ensure_focused();
```
damianavila: I don't thing this line is necessary... selecting the cell by index in the previous should be enough, I think.

damianavila:
Sorry for the delay in the review @parmentelat 
I left some comments. 
If you can not work on this further, just let me know. Thanks!

parmentelat:
Hi

It's been a long time so things are not 100% clear in my memory, but...

I did try to use jupyter's notion of the selected cell at first; 
the code was much simpler indeed, but, in my experience, it's very easy to move inside the slideshow and to *not* have the selected cell - in the jupyter sense - be present in the slice that you are watching;

so I found that relyong on jupyter's selected cell was definitey *not* what I was after most of the time; OTOH since I wrote it this code gives me complete satisfaction

I have some limited availability to work on this, but before you make any decision please try both approaches in real situation

either way, let me know what I can do

thanks for your feedback

damianavila:
>it's very easy to move inside the slideshow and to not have the selected cell - in the jupyter sense - be present in the slice that you are watching;

Interesting...

>so I found that relying on jupyter's selected cell was definitey not what I was after most of the time; OTOH 

OK, now with the context I better get the idea of your PR and I like it!
Sorry to not read your message above... you were clear on it, but I just went to read the implementation :wink:

OK, I will correct any issue in your approach (if I find something) and merge this one.

Thank you!!

rise/static/main.js
damianavila: OK, now testing the thing a little bit more I see why you used this line.

damianavila:
LGTM, merging! Thanks @parmentelat !!

PR #272, opened 2017-06-23
adding autolaunch

choldgraf:
@yuvipanda and I were discussing this today and it didn't seem like it'd take too much code to support auto-launching with metadata.

This is a small patch to add an `autolaunch` metadata parameter, and an extra check at the end of `setup()` that checks for this param and launches RISE if it is `true`.

I'm submitting the PR without testing this extensively to get some feedback on whether this seems like it'll work to you...if so, lemme know what else you think should be added and I can try it out more extensively.

### DIFF
--- a/rise/static/main.js
+++ b/rise/static/main.js
@@ -205,6 +205,20 @@ function setScrollingSlide(config) {
 
 }
 
+/* Setup the auto-launch function, which checks metadata to see if
+*  RISE should launch automatically when the notebook is opened.
+*/
+function autoLaunch(config) {
+
+  var autolaunch_promise = config.get('autolaunch');
+  autolaunch_promise.then(function(autolaunch){
+    if (autolaunch === true) {
+          revealMode();
+      }
+  });
+
+}
+
 /* Setup a MutationObserver to call Reveal.sync when an output is generated.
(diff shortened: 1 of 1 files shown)


choldgraf:
ugh it looks like my editor removed a bunch of trailing whitespaces in the file...lemme know if you want me to revert that

rise/static/main.js
```diff
@@ -619,6 +620,13 @@ function setup() {
     return true;
   };
   $(document).keydown(document_keydown);
+
+  // autolaunch if specified in metadata
+  var config = configSlides()
+  var autolaunch = config.get_sync('autolaunch');
```
damianavila: I would suggest using a pattern like this one: https://github.com/damianavila/RISE/blob/master/rise/static/main.js#L191-L193

You can see there I used a promise... I found that way to be more reliable at the time to load config file values outside the Revealer function.

choldgraf:
So something like this? (see latest push)

choldgraf:
\*sneeze\* oops sorry, apologies for sneezing, didn't mean to distract you, not as if I am trying to ping this thread or anything ;-)

damianavila:
LGTM and tested OK, let's merge it!

PR #276, opened 2017-07-24
Improvements to README

datapythonista:
Not sure if it's just me, but when following the installation instructions in the README, I first thought that after the conda install, it was also necessary to do the `jupyter-nbextension install`. I think changing each option section to be a header 3, makes it more clear.

Also this PR fixes couple of typos in the same file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,13 +8,13 @@
 
 From the most simple to the most complex one, you have 3 options:
 
-1 - Using conda (recommended):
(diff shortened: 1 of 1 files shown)


Isaac1989:
Thanks @datapythonista for making things clearer. I actually thought `jupyter-nbextension install` and others like it was absolutely necessary.

damianavila:
Thanks for the PR, I think this make things more clear. Let's merge it.

Issue #282, opened 2017-08-30
autoexec-fragment ?


parmentelat:
Hi; mostly thinking aloud here, but..

Would it make sense to create a new fragment-type, for example 'autoexec-fragment', that would cause the cell to be executed when the fragment shows up ?

not sure it that makes sense to many people, but I have the feeling that in my case in particular, it would make the whole presenter's life a lot easier; It could in particular contribute smoothening issue #254, that is related in the sense that the only reason why I need to select a cell is so I can evaluate it. 

well again, just a thought - thanks !

damianavila:
In fact I have implemented something to solve #254 for the JupyterCon talk but I have to put it in a PR now.
It essentially auto selects the first cell of each slide.
Regardless of that... maybe we can use the tags availability to mark a fragment as one to be executed when it is shown. I would avoid to add more types of fragments if we can do that with metadata.

Btw, thanks for posting your ideas!

parmentelat:
It's great to hear that you have a solution for #254 and I am eager to be able to use that :)

as for how concretely marking a fragment for auto-execution, as I said I wrote the first idea that came to mind:)
I don't clearly see however what you mean by 'tags availability', can you be more specific about  how this would look like in the metadata ?

damianavila:
> I don't clearly see however what you mean by 'tags availability', can you be more specific about how this would look like in the metadata ?

I am talking about this: https://github.com/jupyter/notebook/pull/2048.
We could potentially read an auto-execute tag from the cell metadata and trigger the execution after the fragment is shown... I think... but I need to think a little bit more about it.

parmentelat:
OK, thanks for making this more clear

Is there an easy way for me to benefit from the patch you made for jupytercon ? 
it'd really help :)

thanks !

damianavila:
The workaround is the following:

Add this piece of code in main.js:

      // select the first cell from each slide
      setTimeout(function(){
        var current_cell_index = reveal_cell_index(Jupyter.notebook);
        console.log(current_cell_index);
        // select and focus on current cell
        Jupyter.notebook.select(current_cell_index)
      }, 500);

after lines:

https://github.com/damianavila/RISE/blob/master/rise/static/main.js#L332

and

https://github.com/damianavila/RISE/blob/master/rise/static/main.js#L338

That will select every first cell of each slide when you change the slides... if 500 ms is too much for you, you can probably diminish it (I tested with 200ms and seems to work OK).

[...]

Issue #280, opened 2017-08-21
Installing development version results in error for reveal


mpacer:
When running 

```
npm run reset-reveal
```

it errs 

```
> rise@0.5.0 reset-reveal /Users/mpacer/jupyter/RISE
> sed -i '11 s_^_/*_' rise/static/reveal.js/css/reveal.css

sed: -i may not be used with stdin

npm ERR! Darwin 16.5.0
npm ERR! argv "/Users/mpacer/anaconda3/envs/dev/bin/node" "/Users/mpacer/anaconda3/envs/dev/bin/npm" "run" "reset-reveal"
npm ERR! node v6.11.0
npm ERR! npm  v3.10.10
npm ERR! code ELIFECYCLE
npm ERR! rise@0.5.0 reset-reveal: `sed -i '11 s_^_/*_' rise/static/reveal.js/css/reveal.css`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the rise@0.5.0 reset-reveal script 'sed -i '11 s_^_/*_' rise/static/reveal.js/css/reveal.css'.
```

mpacer:
This is a known issue with BSD vs. GNU `sed`: 

https://stackoverflow.com/questions/21242932/sed-i-may-not-be-used-with-stdin-on-mac-os-x

The command for OSX is `sed -i '' '11 s_^_\/*_' rise/static/reveal.js/css/reveal.css`

Is there a platform agnostic version that will allow this change?

damianavila:
Closed by #285.

PR #285, opened 2017-09-07
fix for correct invokation of sed on the MAC platform when running

parmentelat:
npm run reset-reveal
as per the development guide
https://github.com/damianavila/RISE#development

this was touched on in #282

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   "scripts": {
     "build-css": "npm run less",
     "build-reveal": "cp -r ./node_modules/reveal.js rise/static/reveal.js",
-    "reset-reveal": "sed -i '11 s_^_\/*_' rise/static/reveal.js/css/reveal.css",
+    "reset-reveal": "sed -i.upstream '11 s_^_\/*_' rise/static/reveal.js/css/reveal.css",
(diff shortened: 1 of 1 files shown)


damianavila:
@mpacer, I think you reported me this personally some days ago.

mpacer:
I have never seen a command line flag like that. No joke, I don't know how to interpret it. But I'll try it in a bit

damianavila:
@mpacer I never saw it before as well :wink:

Here you have some interesting information about it: https://stackoverflow.com/questions/16745988/sed-command-with-i-option-in-place-editing-works-fine-on-ubuntu-but-not-mac/44864004#44864004

mpacer:
sed with an I I've seen (that link is what I used to figure out that it wouldn't work on macOS but what's throwing me off is the `-i.upstream` never seen a dotted command line flag and don't know how to interpret it(even if it's no different)

mpacer:
Oh I see it's a necessary argument to create the backup with that extension. Ok sounds good…

parmentelat:
just to add that this relates to #280 as well

damianavila:
Tested on linux and works. Merging now.

Issue #290, opened 2017-09-12
how to remap Alt-R (was: Is Escape inactive on purpose in reveal mode ?)


parmentelat:
I would suggest that we reinstate the Escape keyboard shortcut for getting out of reveal just like Alt-R does already

Any objection to that ?

parmentelat:
ok, so this was silly, Escape of course means get out of edit mode and go back to command mode, so this would ruin it I guess.

I guess I should rephrase this; Using Alt-R to go in and out of reveal is getting hard on my thumb, so I'd like to ease the pain; I don't seem to be able to map getting in and out of reveal to another keystroke in my `custom.js`, am I missing something

Note: I am editing the title for this issue to reflect this

damianavila:
> Escape of course means get out of edit mode and go back to command mode, so this would ruin it I guess.

That's correct.

>  I don't seem to be able to map getting in and out of reveal to another keystroke in my custom.js, am I missing something

`custom.js`? You probably need to modify `main.js` here: https://github.com/damianavila/RISE/blob/master/rise/static/main.js#L630

parmentelat:
No, I really mean `custom.js`, as I would expect RISE users to be able to do that without having to hack `main.js`; at first I thought that `main.js` would expose some global function that could then be bound to a keyboard shortcut using the usual jupyter tools, but I could not find any such function.

from another angle, I take it it is a standard need from a jupyter extension to be able to bind an extension-related function to a hotkey; there seems to be some elaborate means for an extension to expose itself to the configurator (see attached with a random extension); 

I'm not saying we need to go that far, but definitely being able to at least change that single key seems like a reasonable thing to expect, IMHO

![screen shot 2017-09-12 at 14 50 01](https://user-images.githubusercontent.com/4519227/30326867-39fcda1c-97ca-11e7-9908-05529538c4a9.png)

damianavila:
> No, I really mean custom.js, as I would expect RISE users to be able to do that without having to hack main.js; at first I thought that main.js would expose some global function that could then be bound to a keyboard shortcut using the usual jupyter tools, but I could not find any such function.

There is some discussion to support custom js here: https://github.com/damianavila/RISE/issues/95#issuecomment-89747573

> I'm not saying we need to go that far, but definitely being able to at least change that single key seems like a reasonable thing to expect, IMHO

I agree we need to support a better way to handle shortcut customizations.

parmentelat:
as an update on this issue

## solution 1.

[...]

Issue #114, opened 2015-05-15
hotkeys for slide type


yannpaul:
Does it make sense to have hot keys in command mode to change a cell's slide type. It would be useful in presentation mode, then you wouldn't need the cell toolbar (disabled currently). I propose that one sequence is used to cycle through the types, someting like Alt-s or Alt-t. And the cycling would not include 'skip', so you'd have to jump out of presentation mode or toggle the cell toolbar if that's what you'd like to do.

damianavila:
This is interesting... there is some complexity from the UI perspective about how to select the difference cell types... having shortcuts can help people... but there is some problems because to actually see the changes you have to go out presentation more and enter again... in any case, having shorcuts at "design" time instead of going to the celltoolbar for each cell is interesting.

Alexspayne:
This would be great to have.   Right now the only way to change the slide type is with the mouse, and that is the only thing I need to do with the mouse.  We are really close to being able to do with entirely with the keyboard.

Alexspayne:
It probably makes sense to add this to Jupyter's core however since the metadata dropdown isn't exclusive to the RISE extension.

damianavila:
#291 and #304 added default shortcuts for slide, subslide and fragments. In the future #306 the rest of the cell types will be available + configurable. Closing here since the main request is fulfilled with the current implementation.

PR #291, opened 2017-09-15
keyboard shortcut customization

parmentelat:
a proposal for issue #290


damianavila:
Thanks! Trying to review and merge this one in the next few days.

rise/static/main.js
```diff
@@ -583,6 +579,66 @@ function reveal_cell_index(notebook) {
...
+    actions.register(
+      {
+        help: '(un)set current cell as a Fragment cell',
+        handler: function() {
+          let slideshow = init_metadata_slideshow();
+          slideshow.slide_type = (slideshow.slide_type == 'fragment') ? '' : 'fragment';
+	  Jupyter.CellToolbar.rebuild_all();
+        }
+      },
+      "toggle-fragment", "RISE");
```
damianavila: You have `slide` and `fragment` here... I think we should add the other options as well: subslide, notes and skip.

rise/static/main.js
```diff
@@ -583,6 +579,66 @@ function reveal_cell_index(notebook) {
...
+    actions.register(
+      {
+        help: 'edit all cells (all cells go to edit mode)',
+        handler: function() {
+          Jupyter.notebook.get_cells().forEach(function(cell){
+	    cell.unrender();
+          })
+        }
+      },
+      "edit-all-cells", "RISE");
```
damianavila: In which use cases render-all-cells and edit-all-cells are useful?

rise/static/main.js
```diff
@@ -615,25 +671,36 @@ function revealMode() {
   }
 }
 
+// users can bind another key with e.g.
+// Jupyter.notebook.keyboard_manager.command_shortcuts.set_shortcut('alt-a', 'RISE:slideshow');
+// and likewise bind helper actions with e.g.
+// Jupyter.notebook.keyboard_manager.command_shortcuts.set_shortcut('shift-i', 'RISE:toggle-slide');
+
```
damianavila: Maybe we should set the shortcut by default? There were some request about that in the past: #114

rise/static/main.js
```diff
@@ -615,25 +671,36 @@ function revealMode() {
...
+  Jupyter.toolbar.add_buttons_group([{
+    label   : rise_label,
+    icon    : 'fa-bar-chart-o',
+    callback: revealMode,
+    id      : 'RISE'
+  }]);
+  // define the slideshow action
+  var slideshow_action = {
+    help    : rise_label,
+    handler : revealMode,
```
damianavila: I am seeing this in my box

screen shot 2017-10-31 at 6 45 20 pm

Do you see it as well?

damianavila:
Btw, there is some conflict that prevent this one to merge cleanly.

rise/static/main.js
parmentelat: oh yes, maybe; it would even be possible to factorize all 5, and make this shorter;
I was an a tight schedule at the time though, and am not too available right now either to test this (plus: I  just reinstalled my mac and am still missing a lot of the basic env..)

I can do this in a separate PR if you agree

rise/static/main.js
parmentelat: well, in presentation mode, having a one-key shortcut to so render-all-cells is really convenient for all the cases where you just made a few changes in your cells and forgot to render them properly.

the other way around (edit-all-cells), it is less obvious; I use this much more occasionnally to be honest

rise/static/main.js
parmentelat: well this is really up to you, as I mentioned in the past, chosing the right keyboard shortcut is kind of tricky, and personal too I believe, and I'd rather not waste time discussing these. 
If you are fine with the ones in the comments we can use this of course, people will easily be able to change them later on (was the whole point for this PR in the first place ;)

rise/static/main.js
parmentelat: as I said I don't have this installed on my current machine, so I can't tell you for sure.

I 'm sure I have already seen something similar at some point; I guess this could be fixed by adding some data in slideshow_action, but I don't know how to this off the top of my head

Again if that's a big deal I would be willing to fix this in a subsequent PR

rise/static/main.js
damianavila: > I can do this in a separate PR if you agree

Yes, of course!!

rise/static/main.js
damianavila: Yes, my feeling was the same, render-all-cells seems like something frequently used, but not for edit-all-cells... still, I like the symmetry having both actions, so I am OK to leave it as is.

rise/static/main.js
damianavila: Yes, I think we need to add default values and I am happy with the ones you have in the comments... and as you said, people can change them if they disagree with the choice.

rise/static/main.js
damianavila: > Again if that's a big deal I would be willing to fix this in a subsequent PR

That is OK for me.

parmentelat:

rise/static/main.js
damianavila: I have a branch where I added the subslide action, but I would interested to see some refactoring of this piece so we can offer shortcuts for the rest of the celltoobar options without repeating code.

rise/static/main.js
damianavila: As I said before I will keep this two but I am not assigning them any default shortcut. It is up to the user to configure it if they are interested, at least for now.

rise/static/main.js
damianavila: FYI, I added default values for slide, subslides and fragments in another branch.

rise/static/main.js
damianavila: You were calling the action in the incorrect way, solved with this diff I am adding in the other branch:
```diff
diff --git a/rise/static/main.js b/rise/static/main.js
index f0eec6d..0c3fd49 100644
--- a/rise/static/main.js
+++ b/rise/static/main.js
@@ -802,7 +802,7 @@ function setup() {
   // register action
   Jupyter.notebook.keyboard_manager.actions.register(slideshow_action, "slideshow", "RISE");
   // bind action to keyboard shortcut
-  Jupyter.notebook.keyboard_manager.command_shortcuts.add_shortcut('alt-r', slideshow_action);
+  Jupyter.notebook.keyboard_manager.command_shortcuts.add_shortcut('alt-r', 'RISE:slideshow');

   // same with utility actions
   registerHelperActions();
```

damianavila:
Merging this one now. Thanks @parmentelat for your work here!

Issue #292, opened 2017-09-16
optimizing Shift-Enter


parmentelat:
Still trying to minimize the amount of keystrokes .. ;)

I'm finding that we could still improve user experience, in doing so that Shift-Enter moves to the next cell whenever this one is already displayed

I'm currently playing with this idea, and might have a PR soon

at first sight what I'd like to have is (in addition to always evaluating the cell, of course)

* shift-enter on cells that are last on a slide: do nothing, like now
* shift-enter on cells that are last on their fragment: trigger next fragment
* shift-enter on other regular cells: move to next cell

I'm not quite sure yet how accurately I can manage the fragment thing, we'll see

Now I expect this new behaviour could be linked to some sort of configuration, although I'm not sure how much sense it makes to increase the number of settings. 

What do others think about that idea ?

damianavila:
> Now I expect this new behaviour could be linked to some sort of configuration, although I'm not sure how much sense it makes to increase the number of settings.

I am open to increase the number of settings if it is needed.

> shift-enter on cells that are last on a slide: do nothing, like now

It is important to keep this behavior, otherwise execution will trigger the slide change. And it should be independent events. But you are proposing to keep that, so we are OK.

damianavila:
Closed by #295.

PR #295, opened 2017-09-17
Smart exec

parmentelat:
here's a first rough implementation for #292 

I'm calling it smart exec; it's bound to shift-enter, and allows to execute the current cell and then proceed to the next cell when the context permits

as of this code there is no config flag yet, at this point it is more a base for discussion

### DIFF
--- a/rise/static/main.js
+++ b/rise/static/main.js
@@ -137,7 +137,6 @@ function markupSlides(container) {
     for (var i=0; i < cells.length; i++) {
         var cell = cells[i];
         var slide_type = get_slide_type(cell);
-        //~ console.log(`cell ${i} is: ${slide_type}`);
 
         if (content_on_slide1) {
             if (slide_type === 'slide') {
@@ -182,7 +181,6 @@ function markupSlides(container) {
             cell.element.addClass('reveal-skip');
         }
 
-      //console.log(`marked cell ${i} as ${cell.smart_exec}`)
     }
 
   // set on all cells a smart_exec tag that says how smart exec
--- a/rise/static/main.js
+++ b/rise/static/main.js
@@ -138,7 +138,7 @@ function markupSlides(container) {
     function set_smart_exec_based_on_next(cells, i) {
       var cell = cells[i];
       // default is 'pinned' because this applies to the last cell
-      var tag = 'pinned';
+      var tag = 'smart_exec_slide';
       for (var j = i+1; j < cells.length; j++) {
         var next_type = get_slide_type(cells[j]);
         if ((next_type == 'slide') || (next_type) == 'subslide') {
--- a/rise/static/main.js
+++ b/rise/static/main.js
@@ -134,29 +134,7 @@ function markupSlides(container) {
 
     var cells = Jupyter.notebook.get_cells();
 
-    // implement the 'smart_exec' tag; ignore notes and skip and the like
-    function set_smart_exec_based_on_next(cells, i) {
-      var cell = cells[i];
-      // default is 'pinned' because this applies to the last cell
-      var tag = 'smart_exec_slide';
-      for (var j = i+1; j < cells.length; j++) {
-        var next_type = get_slide_type(cells[j]);
-        if ((next_type == 'slide') || (next_type) == 'subslide') {
-          tag = 'smart_exec_slide';
-          break;
-        } else if (next_type == 'fragment') {
-          tag = 'smart_exec_fragment';
-          break;
-        } else if (next_type == '') {
-          tag = 'smart_exec_next';
-          break;
-        }
-      }
-      cell.smart_exec = tag;
-    }
-
-    for (var i=0; i < cells.length; i++) {
-        var cell = cells[i];
+    for (var cell of cells) {
         var slide_type = get_slide_type(cell);
         //~ console.log(`cell ${i} is: ${slide_type}`);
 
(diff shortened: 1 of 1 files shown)


parmentelat:
I had recorded a session as a gif, but it resulted in a file larger than 10mb, so..

the points I wanted to make in this recorded session were as follows

* all this makes sense only with auto_selection turned on, and that's not merged yet
which is also why I have not implemented a config flag yet

* when doing shift-enter on a cell X that is the last one in its fragment, right now it is stuck (i.e. shift enter does not move to the next cell); however is the case where the following fragment has been shown and we have moved the selection back on cell X, one could argue that shift-enter should then indeed move to the next cell. 

  this IMO would be the ideal behaviour but it is
   * of relative minor importance, and 
   * significantly harder to implement.

  And so I ruled it out for now.

parmentelat:
I found a way to get this to work the way I described as ideal above, so I changed my mind and went for it

parmentelat:
here's what it looks like - this requires auto-select and auto-select-fragment

rise-smart-exec2

damianavila:
This is really exciting, thanks for the gif.

>all this makes sense only with auto_selection turned on, and that's not merged yet
which is also why I have not implemented a config flag yet

No problem... I need to test, review and merge the other PR. Hopefully I can do that in the next few days.

>I found a way to get this to work the way I described as ideal above, so I changed my mind and went for it

+1

> this requires auto-select and auto-select-fragment

I have not tested/reviewed yet, but how is the general behavior without these requirements? I am willing to make those a default thing if we can provide a unified experience. I would not like to have the smart-exec *only* when some specific options are configured.

parmentelat:
To clarify:
The 2 features are vastly independent in terms of their implementation
 I was just saying that the gif had been recorded with auto-select turned on as well

damianavila:
> I was just saying that the gif had been recorded with auto-select turned on as well

Ah... OK, thanks for the clarification.

rise/static/main.js
```diff
@@ -77,8 +77,30 @@ Object.getPrototypeOf(Jupyter.notebook).get_cell_elements = function () {
...
+function get_slide_type(cell) {
+  var slide_type = (cell.metadata.slideshow || {}).slide_type;
+  return ( (slide_type === undefined) || (slide_type == '-')) ? '' : slide_type;
+}
+
+function is_slide(cell)    {return get_slide_type(cell) == 'slide';}
+function is_subslide(cell) {return get_slide_type(cell) == 'subslide';}
+function is_fragment(cell) {return get_slide_type(cell) == 'fragment';}
+function is_regular(cell)  {return get_slide_type(cell) == '';}
+
```
damianavila: We should be reusing these in other places, but that's for the future...

rise/static/main.js
```diff
@@ -111,12 +133,11 @@ function markupSlides(container) {
...
-    var i, cell, slide_type;
 
-    for (i=0; i < cells.length; i++) {
-        cell = cells[i];
-        slide_type = (cell.metadata.slideshow || {}).slide_type;
-        //~ console.log('cell ' + i + ' is: '+ slide_type);
+    for (var i=0; i < cells.length; i++) {
+        var cell = cells[i];
+        var slide_type = get_slide_type(cell);
+        //~ console.log(`cell ${i} is: ${slide_type}`);
```
damianavila: This comment can be removed.

rise/static/main.js
```diff
@@ -129,15 +150,18 @@ function markupSlides(container) {
                 // Start new subslide
                 current_fragment = subslide_section = new_subslide();
             } else if (slide_type === 'fragment') {
-                current_fragment = $('<div>').addClass('fragment')
+              // record the <div class='fragment'> element corresponding
+              // to each fragment cell in the 'fragment_div' attribute
```
damianavila: Why you need to do this? I guess the answer is below.

rise/static/main.js
```diff
@@ -157,6 +181,41 @@ function markupSlides(container) {
         if (slide_type === 'skip') {
             cell.element.addClass('reveal-skip');
         }
+
+      //console.log(`marked cell ${i} as ${cell.smart_exec}`)
```
damianavila: Another debug thing that can be removed.

rise/static/main.js
```diff
@@ -157,6 +181,41 @@ function markupSlides(container) {
...
+    }
+
+  // set on all cells a smart_exec tag that says how smart exec
+  // should behave on that cell
+  // the fragment cell also get a smart_exec_next_fragment
+  // attribute that points at the <div class='fragment'>
```
damianavila: >Why you need to do this? I guess the answer is below.

And here is the answer :wink:

rise/static/main.js
```diff
@@ -554,14 +640,6 @@ function reveal_cell_index(notebook) {
   // just scan all cells until we find one at that address
   // except that we need to start at -1 0r 0 depending on
   // whether the first slide has a slide tag or not
-  var is_slide = function(cell) {
```
damianavila: Nice to see we are removing duplication.

parmentelat:
I have taken care of the extra debug statements

As far as the conflict: this page here https://github.com/damianavila/RISE/pull/295/conflicts
shows me one conflict on line 697; but I don't understand how this is a conflict; the 3 old function definitions must be removed from there indeed..

I don't see how I can help out from my end, let me know if you need more info to merge this

damianavila:
Try merging master into your branch... that should fix the issue, I think.

parmentelat:

damianavila:
> I need to go for today; let me know if you need anything else on my end for merging this

I think we are OK, thanks for your feedback.

damianavila:
Btw, #308 add some comments about the functionality in the usage docs.

Issue #173, opened 2015-12-15
Adding header or footer to slides


nigeljyng:
Is there any way to add a header or footer to every slide? Other than hacking the source code.

Thanks for the great work, Damian.

damianavila:
pinging @fpliger :wink: he did some work on this, but not sure if he abstracted the thing enough to make a PR (I hope so...)

fpliger:
Ah.. I'm guilty on this. :)

I did some work that [I guess] can be reused and [hopefully] turned into a PR. @nigeljyng what are your time constraints? I'm on many different deadlines ATM... :/

nigeljyng:
Early January would be nice, if that's feasible for you!

jmansour:
Hey folks.  End up getting any traction on this feature?

jneines:
I would like to give a +1 on this. I'm looking into using RISE as another option for company level presentations and therefore need to follow corporate identity design. It's important to be able to place logos, Author, company and date at a specific location visible throughout the presentation. Furthermore a section title should be visible at the top of the screen as well. This definitly sounds like using header and footer areas for this.

I'm currently looking into using a custom.js to relocate header elements (h1,h2,..) out of the slides context to make this happen via jquery based insertBefore and insertAfter, but with limited success. Is there a RISE specific javascript event, which I could hook on to when slideshow mode is enabled? Does the idea sound reasonable after all?

damianavila:
> s there a RISE specific javascript event, which I could hook on to when slideshow mode is enabled? 

You can probably use some of the Reveal.js events... or maybe use some of the notebook machinary to signal that...

>Does the idea sound reasonable after all?

It is one approach, probably not the one I would choose. @fpliger do you have the branch you were working on this doing experimentation? It would be nice to make it public so others can continue the work :wink: Thanks!

fpliger:
>> Does the idea sound reasonable after all?

> It is one approach, probably not the one I would choose. @fpliger do you have the branch you were working on this doing experimentation? It would be nice to make it public so others can continue the work 😉 Thanks!

Oh man.. I've changed laptop in the meanwhile. Will check but not feeling very lucky there.

damianavila:
> Oh man.. I've changed laptop in the meanwhile. Will check but not feeling very lucky there.

Thanks OK...let's see if we are lucky... otherwise if you can drop the ideas (if you remember them). Thanks!

fpliger:
@damianavila  found something... 😄 

[...]

PR #296, opened 2017-09-21
css loading, header and footer: a proposal

parmentelat:
This commit adds a couple features and related examples

* css-loading (inspired by PR #200, a little extended), in that order

  * ./rise.css (so this will apply to all notebooks in the current dir)
  * ./<notebookname>.css - this namely is what PR #200 was about

* header and footer

  * looks for config tag 'overlay' - if found, this is used
    to build the constant background;
    it is wrapped in a <div>, so can be text or html
    in this case user is entirely responsible for styling

  * otherwise looks for config tags 'header' and 'footer'
    in this case minimum styling is applied (floor and ceiling)
    but user is still responsible for cosmetic styling

See the examples directory for usages
I've taken that chance to add an example named showflow
to demonstrate other features related to controlling
the slideshow with space and shift-enter
this is unrelated but a harmless addition imho

Relates to:

#173 for header and footer
#200 for adding css loading


damianavila:
Wonderful @parmentelat, I would love to have more time to give you feedback on these PR more quickly. Hopefully I can review, test and merge several of them at the weekend. Sorry for the delay and thank you very much for your multiple contributions!

parmentelat:
> Wonderful @parmentelat, I would love to have more time to give you feedback on these PR more quickly. Hopefully I can review, test and merge several of them at the weekend. Sorry for the delay and thank you very much for your multiple contributions!

not at all, thank *you* for putting this together in the first place :)

I’m rushing to get ready before we start our video recording session, that’s why I am frantically trying to get this to work as smoothly as I can, and that’s why you’re getting this flood of changes

rise/static/main.js
```diff
@@ -248,6 +248,39 @@ function disconnectOutputObserver() {
   }
 }
 
+function addHeaderFooterOverlay(config) {
+    let overlay = config.get_sync('overlay');
+    let header =  config.get_sync('header');
+    let footer =  config.get_sync('footer');
+    let backimage =  config.get_sync('backimage');
```
damianavila: I think we need to put default values here https://github.com/parmentelat/RISE/blob/master/rise/static/main.js#L25, for these options. Don't you think?

rise/static/main.js
```diff
@@ -275,6 +308,18 @@ function Revealer(selected_slide, config) {
...
+  /* this policy of trying ./rise.css and then <notebook>.css
+   * should be redefinable in the config
+   */
+  if (window.location.pathname.endsWith('.ipynb')) {
+      // Attempt to load rise.css 
+      $('head').append(`<link rel="stylesheet" href="./rise.css" id="rise-custom-css" />`);
+      // Attempt to load CSS with the same path as the .ipynb but with .css extension instead
+      let notebook_css = window.location.pathname.replace(/\.ipynb$/,'.css');
+      $('head').append(`<link rel="stylesheet" href="${notebook_css}" id="notebook-custom-css" />`);
+
```
damianavila: Makes sense, you have a rise.css that could be more general applied first and then the notebook specific one applied on top of that...

damianavila:
Some things:

* We need to document how we can use these, I know the examples are good to know how it works, but we need, at least, to put some lines in the config section in the docs as well.

* Are we allowed to use that background image? I don't not know the copyright for that image. If we don't have it clear, we need to use another public image.

rise/static/main.js
parmentelat: well, I'm not quite sure now, but I think at the time I considered it but rules it out because of the config logic; IIRC having overlay defined kind of hides the ability to use header/footer/backimge (or it it the other way around?)

So probably having them in a comment could be helpful, but more than that I don't know

parmentelat:
* generally speaking, it would make sense to rethink the doc entirely IMHO
so I had some notes on how to go about that, but honestly would appreciate if all the code could settle at first

* about the image;l that's the background I was playing with at the time for our video recording; both logos are public, so I would tend to think yes; however I cannot produce a formal agreement from both institutions, so if you want to play it safe please replace with any other image that you'd like

parmentelat:

rise/static/main.js
damianavila: OK, a comment sound like a good compromise.

damianavila:
> it would make sense to rethink the doc entirely IMHO
so I had some notes on how to go about that

Great, would love to hear those thoughts.

>but honestly would appreciate if all the code could settle at first

That's ok. We can do that :wink:

> if you want to play it safe please replace with any other image that you'd like

OK, I will probably do that to play safe.

damianavila:
> I need to go for today; let me know if you need anything else on my end for merging this

I think we are OK, have a fun rest of your day.

damianavila:
Adding some more minor pieces in #310.

PR #311, opened 2017-11-03
now is maybe a right time to fix #293

### LINKED ISSUE #293: coding standards, esp. in `main.js`
I'm a little hindered by the code layout, especially in `main.js`
there are places with 2 spaces indent, some with 4, plus the whole body is not indented inside the main `define` call.

I'm not being dogmatic here, but in practical terms I find that I keep having to fight against emacs that wants to align things differently than the rest of the code, and this is becoming a hassle

I'd like to propose a big cleanup in this area, although I understand it's a pain wrt pending PRs, but a bit of planning could maybe make it possible ?

thoughts ?

parmentelat:
I figured, if we dont do this now, we wont :)
I have done this as carefully as possible, but have still no way to actually test this.
I have not changed the code at all except for formatting
It makes a huge amount of changes, but if you ignore whitespaces it's quite reasonable to review.

Please let me know if you plan on accepting this or not (no worries if not, I understand it's quite intrusive), and if yes on what sort of timeline.

### DIFF
--- a/rise/static/main.js
+++ b/rise/static/main.js
@@ -1,130 +1,136 @@
 /* -*- coding: utf-8; js-indent-level: 2 -*-
-* ----------------------------------------------------------------------------
-* Copyright (c) 2013-2017 Damián Avila and contributors.
-*
-* Distributed under the terms of the Modified BSD License.
-*
-* A Jupyter notebook extension to support *Live* Reveal.js-based slideshows.
-* -----------------------------------------------------------------------------
-*/
+ * ----------------------------------------------------------------------------
+ * Copyright (c) 2013-2017 Damián Avila and contributors.
+ *
+ * Distributed under the terms of the Modified BSD License.
+ *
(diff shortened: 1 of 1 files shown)


damianavila:
This is nice, but I would prefer to have the release with the current state and, immediately after the release, just merge this out and keep iterating, if necessary, with more PRs. In that way we minimize the risks associated with these changes from impacting the release. Hope it sounds reasonable to you. Thanks for pushing on this!!

parmentelat:
that's fine with me; let me known when the release is out
I'm eager to be able to use all that stuff from a clean install :)

parmentelat:
so now that the release is out, how about merging this ?

I could use some spare cycles on this project, but in the current state of affairs I'm reluctant to start any change in `main.js` that could create a lot of conflicts with this one change

damianavila:
Let's merge it and iterate! And very happy to see you pushing for more contributions!!

PR #318, opened 2017-11-17
Adding customization reference

choldgraf:
This PR does 2 things:

1. Deletes `_build` (that shouldn't be in git history) and adds it to `.gitignore`. I recommend removing this folder entirely from git history but will leave that to someone else if desired.
2. Adds a list of configuration parameters and links to their section in the documentation. See here:

http://predictablynoisy.com/RISE/customize.html#configuration-options

There are a few options at the end that aren't documented, but again I'll leave that to another PR. This lays the groundwork to make that addition relatively easy in the future.


choldgraf:
ping

damianavila:
pong :wink:

Will review this week. Btw, thanks for the PR and for the pings (they are useful to find myself time even when there is not free boxes in the schedule). Thanks!

damianavila:
> Deletes _build (that shouldn't be in git history) and adds it to .gitignore. I recommend removing this folder entirely from git history but will leave that to someone else if desired.

Mmm... I have doubts about it. I can break people if I rewrite history like this. 
@choldgraf @parmentelat, more thoughts about this?? (pinging you because you would be affected by this pruning)

In any case, this is not something blocking this PR.
And LGTM, so I am going to merge it and deploy it.
Thanks @choldgraf.

choldgraf:
cool! thanks!

PR #320, opened 2017-11-19
Made the toolbar button narrower, now it only says 'RISE Slideshow'

### LINKED ISSUE #319: The toolbar button is very wide
The _Enter/Exit Live Reveal Slideshow_ button in the toolbar is very wide.

screen shot 2017-11-19 at 21 53 18

Being a button, it already implies a command. Further, the button is never actually used for exiting, only entering the slideshow.

This is a cosmetic (ie. important) issue, and becomes more of a functional issue when the user is installing other extensions which add buttons to the toolbar.

xmacex:
Here is a narrower toolbar button, fixes #319 .

screen shot 2017-11-19 at 22 15 33

Should it be still leaner? With maybe just the word "Slideshow" on it, or a verb like "Present", or just the icon, coherent with most of the other buttons? Personally this button has become important to me (thank you for amazing extension!), so I would enjoy having a word in the button.

### DIFF
--- a/rise/static/main.js
+++ b/rise/static/main.js
@@ -928,19 +928,16 @@ define([
   function setup() {
     $('head').append('<link rel="stylesheet" href=' + require.toUrl("./main.css") + ' id="maincss" />');
 
-    // use same label in button and shortcut
-    var rise_label = 'Enter/Exit RISE Slideshow';
-
     // create button
     Jupyter.toolbar.add_buttons_group([{
-      label   : rise_label,
+      label   : 'RISE Slideshow',
       icon    : 'fa-bar-chart-o',
       callback: revealMode,
       id      : 'RISE'
     }]);
     // define the slideshow action
     var slideshow_action = {
-      help    : rise_label,
+      help    : 'Enter/Exit RISE Slideshow',
       handler : revealMode,
     }
     // register action    


damianavila:
Yes, it is too long... I would leave just RISE, I think.

dsblank:
FYI, I think "slideshow" is more informative than "RISE" (which means
nothing to many users, even after they know what the functionality). From a
UX perspective, "slideshow" makes more sense; from a marketing perspective,
sure, use the product name.

parmentelat:
all other extensions use a single icon as the button message;

we used to have just one icon in the toolbar as well, and that was fine IMHO; remember there is a pop-up legend when you hover the mouse on the button, and right now this properly says `Enter-Exit RISE Slideshow`.

I'm not quite sure how this change made it to the published version, but if I was involved it was clearly unintentional :)

damianavila:
> we used to have just one icon in the toolbar as well, and that was fine IMHO; remember there is a pop-up legend when you hover the mouse on the button, and right now this properly says Enter-Exit RISE Slideshow.

Well, that's true... we have the pop up legend... so I would probably be OK with keeping the icon.

> I'm not quite sure how this change made it to the published version, but if I was involved it was clearly unintentional :)

There was a change upstream in the notebook package that allowed this ;-)
I did realize the change too late in the release process, but we can fix it in the next one.

damianavila:
After thinking about this I am happy the icon + the popup.
BTW, @xmacex there is [PR](#324) from another contributor (@parmentelat) solving this issue in a more fundamental way (and refactoring some other things). Sorry about this by I will close your PR on favor of the other one.
Thank you very much for your contribution, it started a nice discussion and a nice resolution. I hope to see you again submitting PRs/issues. 
THANK YOU!

xmacex:
No problem at all, thanks @damianavila for awesome and useful plugin, and @parmentelat for a finer contribution :) Cheers!

Issue #321, opened 2017-11-23
How about adopting default: auto_select = 'code'


parmentelat:
it's all in the title

do we have any specific reason to stick with the current default that has auto-select basically turned off ? 

I'm asking because the presentations that I can see on binder generally have not bothered to set this otherwise obscure setting, and I feel like it's a bit of a pity

what do others think ?

damianavila:
I think we could probably make auto_select defaults to true. Let's wait a little bit for others and if we don't have comments, we do it.

Additionally, I think we can discuss about other options that needs to to have other values by default, such as, `start_slideshow_at` (I think the default behavior should be "selected").

Is there any other we want to discuss about?

parmentelat:
+1 for defaulting `start_slideshow_at` to `selected`

parmentelat:
added a line on gitter so people have a chance to discuss this

damianavila:
> added a line on gitter so people have a chance to discuss this

Thanks! Let's wait until the end of the week, and then we proceed with the changes if there is not -1.

parmentelat:
just thinking aloud mostly but, would it make sense to also default autolaunch to true whenever a notebook has the 'livereveal' metadata chunk ?

parmentelat:
to summarize the current proposal:

* `start_slideshow_at` would default to `'selected'` (agreement so far)
* `auto_select` would default to `'code'` (agreement do far)
* `autolaunch` could default to `true` if the notebook metadata has a non-empty `livereveal` chunk  (just proposed)

damianavila:
mmm... I am not sure about autolaunch... I think is too much disruptive. I would be -0.5 on that one...

parmentelat:
fair enough, as I said that was just an idea;
let's forget about that then, maybe some other time

damianavila:
> let's forget about that then, maybe some other time

Maybe, with time, we can evaluate the usage for that option and how spread it is and with that info in mind we can revisit the idea.

damianavila:
OK, I merged the PR, let's close this one now.

PR #323, opened 2017-11-30
an implementation for #321

parmentelat:
set defaults as per #321:

      start_slideshow_at: 'selected',
      auto_select: 'code',

### DIFF
--- a/rise/static/main.js
+++ b/rise/static/main.js
@@ -33,14 +33,25 @@ define([
       theme: 'simple',
       transition: 'linear',
       slideNumber: true,
-      start_slideshow_at: 'beginning',
-      /* can be either
-       * "none" - no autoselect,
-       * "first" - select first cell
-       * "code" - select first code cell
+      /* describe where to start slideshow
+       * can be either:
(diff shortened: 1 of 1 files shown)


rise/static/main.js
```diff
@@ -33,14 +33,25 @@ define([
...
-      start_slideshow_at: 'beginning',
-      /* can be either
-       * "none" - no autoselect,
-       * "first" - select first cell
-       * "code" - select first code cell
+      /* describe where to start slideshow
+       * can be either:
+       * 'beginning' : start on first slide 
+       * 'selected'  : start on current slide
+       * xxx: could be useful to add more policies
```
damianavila: sounds like a good idea for the future

damianavila: In fact, for that case, I mean a hard-wired slide number, maybe `reveal.js` itself offers that possibility, and it would be probably very easy to adapt/implement it in RISE

damianavila:
LGTM, merging now since we did not have any other feedback in channels nor the issue.

Issue #306, opened 2017-11-03
Refactor keyboard shortcuts 


damianavila:
* Add `notes` and `skip` into the keyboard actions
* Make the rest of the keyboard shortcuts configurable as well (slideshow view)

parmentelat:
I think this issue should be closed

damianavila:
You are right!

PR #324, opened 2017-11-30
an implementation for #306 and that fixes #319 as well

### LINKED ISSUE #319: The toolbar button is very wide

parmentelat:
this is mostly a big cleanup that uses jupyter's actions mechanism for everything including the toolbar button

once I was done I realized it also fixed #319, in a radically different way, I guess it will obsolete #320, sorry about that, it was not the intention


rise/static/main.js
```diff
@@ -806,62 +818,56 @@ define([
...
-        help: 'edit all cells (all cells go to edit mode)',
-        handler: function() {
-          Jupyter.notebook.get_cells().forEach(function(cell){
-	    cell.unrender();
-          })
-        }
-      },
-      "edit-all-cells", "RISE");
+    // newtype can be any of 'slide' 'subslide' 'fragment' 'note' 'skip'
+    function toggle_slide_type(newtype) {
```
damianavila: this should be `new_type` OR change those references below to `newtype`, otherwise you will get an error...

damianavila:
> this is mostly a big cleanup that uses jupyter's actions mechanism for everything including the toolbar button

Great!

> once I was done I realized it also fixed #319, in a radically different way, I guess it will obsolete #320, sorry about that, it was not the intention

I know, and I think your solution is better so I decided to close the other PR giving the author all my gratitude for their contribution (thanks again @xmacex).

rise/static/main.js
parmentelat: good catch, thanks for spotting that one - fixed in 72c8586

parmentelat:
apologies and thanks to @xmacex from my end as well :)

parmentelat:
Hi Damian

I think I have fixed the issue that you pointed out and so as far as I am concerned this could be merged at your best convenience

let me know if I'm missing something


-- Thierry

damianavila:
> let me know if I'm missing something

Just one more thing and we will merge it: https://github.com/damianavila/RISE/blob/master/doc/customize.rst

Please update that section with the changes we introduced here (pretty sure there is some changes needed around the default value we used to have).

parmentelat:
Hi Damian

I have reviewed `customize.rst` as per your request, and made one last change in `main.js` so that now `smart-exec` is also registered as a jupyter action, which plays better with jupyter's helper tools

let me know if you request anything else before merging this

damianavila:
Changes LGTM, let merge it and iterate if needed. Thanks again for you contributions!

PR #327, opened 2017-12-03
Make the output observer aware of the scrolling needs

damianavila:
Fixes the problem outlined here: https://github.com/binder-examples/jupyter-rise/issues/1

cc @choldgraf

### DIFF
--- a/rise/static/main.js
+++ b/rise/static/main.js
@@ -299,11 +299,12 @@ define([
    * This fixes issue #188: https://github.com/damianavila/RISE/issues/188
    */
   var outputObserver = null;
-  function setupOutputObserver() {
+  function setupOutputObserver(config) {
     function mutationHandler(mutationRecords) {
       mutationRecords.forEach(function(mutation) {
         if (mutation.addedNodes && mutation.addedNodes.length) {
           Reveal.sync();
+          setScrollingSlide(config);
         }
       });
     }
@@ -494,7 +495,7 @@ define([
               });
 
               // Sync when an output is generated.
-              setupOutputObserver();
+              setupOutputObserver(config);
 
               // Setup the starting slide
               setStartingSlide(selected_slide, config);
--- a/rise/static/main.js
+++ b/rise/static/main.js
@@ -256,6 +256,7 @@ define([
         // Start from the beginning
         Reveal.slide(0, 0);
       }
+      setScrollingSlide(config);
     });
 
   }


damianavila:
@choldgraf if you can validate that this is working for you, it would be great!
My tests were successful, so probably merging soon. Thanks!

choldgraf:
hmmm, I'm actually having a hard time getting scrolling to work on my machine. See gif below. Does that configuration look right to you? (it's kinda hard to tell but I'm trying to scroll up/down after the plot is made and nothing happens...

rise

damianavila:
mmm... can you make several cells in one slide to see if the scrolling appears (without the execution part, just several cell). Can you also tell me which browser and version you are using?

Another thing... can you try hard-clearing the cache and reloading to be sure there is not any cache issue involved?

Thanks.

choldgraf:
ah ok - so if I have multiple cells that make it *way* too big for the y-axis then the scrolling does work. So maybe I'm just not getting your patch working properly. Do I have to do something special for RISE to run from editable mode other than `pip install -e RISE`? I know that'll make python recognize the custom install but does that work for javascript too?

damianavila:
> Do I have to do something special for RISE to run from editable mode other than pip install -e RISE? I know that'll make python recognize the custom install but does that work for javascript too?

See here the dev section: https://damianavila.github.io/RISE/dev/develop.html
Particularly, after `pip install -e .`, you need to:
```
jupyter-nbextension install rise --py --sys-prefix --symlink
jupyter-nbextension enable rise --py --sys-prefix
```

Notice the --symlink option... it is there so any JS change get loaded.

choldgraf:
hmmm, I ran those but it doesn't seem to be auto-creating the scroll when the output goes off the bottom of the screen :-/

how do you think I should best-debug this?

choldgraf:
ok actually I found a weirder bug that may be related so some stuff reported here. If I open the notebook, then run the cells, then the scroll bar works as expected. However, if I then exit the slideshow and re-enter it, then it no longer works:

rise

note that this happens on both chrome and firefox (running on windows 10)

damianavila:
>how do you think I should best-debug this?

Try with a console.log after line 307, to see if we arrive to that part of the code...

My best guess is around an issue with the cache... or actually the changed not being applied somehow.

damianavila:
> ok actually I found a weirder bug that may be related so some stuff reported here. If I open the notebook, then run the cells, then the scroll bar works as expected. However, if I then exit the slideshow and re-enter it, then it no longer works:

Can you share with me the ipynb file, and the browser and versions you are using?

choldgraf:
check out: https://mybinder.org/v2/gh/choldgraf/sandbox/master?filepath=rise_demo.ipynb :-)

choldgraf:
(and you are right, on my machine the console.log didn't turn up anything so I dunno why it's not being installed, but I think this bug is separate from that)

damianavila:
> I think this bug is separate from that

That's correct, I can see that behavior, I need to explore this new issue a little more but it is different from the original one.

> (and you are right, on my machine the console.log didn't turn up anything so I dunno why it's not being installed

Yep, you can try installing without symlink option: jupyter-nbextension install rise --py --sys-prefix, which means every time you change the JS, you will need to run that command again.

damianavila:
> That's correct, I can see that behavior, I need to explore this new issue a little more but it is different from the original one.

Easy fix... will add it in this PR as well in a few minutes.

damianavila:
>Easy fix... will add it in this PR as well in a few minutes.

Done, just one line.
If you can not test the PR, I can probably build a conda package for you with this change included... let me know. And thanks for the reports!!

choldgraf:
ah hah - I got it working. I think that there was another rise lingering somewhere that was being unintentionally loaded. Just tested and it worked well!

damianavila:
> ah hah - I got it working. I think that there was another rise lingering somewhere that was being unintentionally loaded. Just tested and it worked well!

Perfect, let's get it merged.

Issue #309, opened 2017-11-03
make RISE configurable with jupyter_nbextensions_configurator


parmentelat:
[jupyter_nbextensions_configurator](https://github.com/Jupyter-contrib/jupyter_nbextensions_configurator) would be a great way to let people configure RISE without needing to mess too much with config files and the like

I was precisely talking on that repo earlier today, so I took that chance to ask for guidance, see this thread here where Josh gave a lot of insights
https://github.com/Jupyter-contrib/jupyter_nbextensions_configurator/issues/49

this would mostly involve writing a [yaml config file](https://github.com/Jupyter-contrib/jupyter_nbextensions_configurator/blob/c48a0b579ba82de467e8e4bbb207882d83864637/README.md#yaml-file-format)

an option to consider at least, IMHO

damianavila:
Open to establish interoperability with the configurator.

parmentelat:
capturing 2 examples of interest, from Josh's post, for extension `collapsible_headings`


* [the yaml file](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/5646a5b75467830e7fef739a42d6a7e88ac9f1a5/src/jupyter_contrib_nbextensions/nbextensions/collapsible_headings/collapsible_headings.yaml#L61-L88) 

* [the js code](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/5646a5b75467830e7fef739a42d6a7e88ac9f1a5/src/jupyter_contrib_nbextensions/nbextensions/collapsible_headings/main.js#L882-L890) that activates the keybindings when enabled

which gives this - still for collapsible headings of course

![screen shot 2017-11-04 at 00 14 09](https://user-images.githubusercontent.com/4519227/32399299-2d0de7d4-c0f5-11e7-8dde-111291eac2ac.png)

ewankirk:
This would be a very useful addition since a lot of people use the nbextensions_configurator.  I am trying to move our users from Powerpoint to RISE for technical presentations but asking them to edit json config files is probably too much.

parmentelat:
Just a heads up: I have started to dust this off, and am planning on doing some progress in the near future. I have several comments and questions about that though:

* there is a need to ship 2 additional files with the installation; I have named them `rise.yaml` and `rise.png` in my current working tree, and they are located in `rise/static/` just like `main.js`; they need to be installed in the same directory as `main.js`; how should I go about doing that (for now I am moving forward just just manually created symlinks..)

[...]

PR #344, opened 2018-02-14
making rise compliant with nbextensions_configurator

parmentelat:
this is a first attempt, reasonably complete as far as I can tell, at solving #309

please make sure to review the commit messages for additional details

I am sorry that I had to make so many changes in `main.js`, I hope it will still be OK for you to review.

Thierry


damianavila:
Hi Thierry, thanks for the PR.

 > I am sorry that I had to make so many changes in main.js, I hope it will still be OK for you to review.

It is OK but it would take me some time to review and test it. Just wanted to let you know.

Cheers.

parmentelat:
of course, and I understand

the one thing that I would appreciate if you could start with it, is about the packaging of the new pieces; I am not very comfortable with this packaging/building business, so I will need some guidance from you on that one, so that we can have this out of the way first thing.

damianavila:
>  I would appreciate if you could start with it, is about the packaging of the new pieces

OK, that would be the first thing I am going to check. Thanks for the additional context.

rise/static/configurator.md
```diff
@@ -0,0 +1 @@
+# This is a placeholder for a markdown file that is displayed by nbextensions_configurator
```
damianavila: Do you want to use this placeholder for a sort of README?

rise/static/main.js
```diff
@@ -13,82 +13,104 @@ define([
...
+   * 1) start from the hardwired settings (in this file)
+   * 2) add the settings from nbextensions_configurator (i.e. .jupyter/nbconfig/notebook.json)
+   *    they should all belong in the 'rise' category
+   * 3) and finally add the settings from the notebook metadata
+   *   3a) for legacy reasons: use the 'livereveal' key
+   *   3b) for more consistency, then override with the 'rise' key
+   * 4) also stores in complete_config.is_slideshow a boolean
+   *    that says if the notebook metadata contains 
+   *    either livereveal or rise
+   *
```
damianavila: What about the config coming from `livereveal.json`?

That is loaded here in https://github.com/damianavila/RISE/pull/344/files#diff-baac928f853273e00536869aeabb2e89L64

IMO, that should be the second step... if we do not add it somehow here we are probably breaking people using that config file to load their config options.

rise/static/main.js
```diff
@@ -13,82 +13,104 @@ define([
...
+      history: true,
       scroll: false,
       center: true,
-      autolaunch: false,
+      margin: 0.1,
+      minScale: 1.0, // we need this for codemirror to work right
     };
 
-    var config_section = new configmod.ConfigSection(
-      'livereveal',
```
damianavila: See my comment above about `livereveal.json`.

rise/static/main.js
```diff
@@ -13,82 +13,104 @@ define([
...
-
-    return final_config
-
+    // 1) initialize with hardwired defaults
+    $.extend(true, complete_config, hardwired_config);
+
+    // this is a ConfigSection object as per notebook/static/services/config.js 
+    let nbext_configurator = Jupyter.notebook.config;
+    // returning this will allow further chaining of then
+    return nbext_configurator.loaded.then(
```
damianavila: So you are waiting here for the notebook config to be loaded to chain it the other events ($.extend)?

rise/static/main.js
```diff
@@ -255,68 +277,61 @@ define([
...
-
-    var autolaunch_promise = config.get('autolaunch');
-    autolaunch_promise.then(function(autolaunch){
-      if (autolaunch === true) {
-        revealMode();
-      }
-    });
-
+  function autoLaunch() {
+    if (complete_config.autolaunch && complete_config.is_slideshow)
```
damianavila: Why are you waiting here for `complete_config.is_slideshow`?

rise/static/main.js
```diff
@@ -420,24 +435,14 @@ define([
...
-              var options = {
-                // All this config option load correctly just because of require-induced delay,
-                // it would be better to catch them from the config.get promise.
-                controls: config.get_sync('controls'),
-                progress: config.get_sync('progress'),
-                history: config.get_sync('history'),
+              // all these settings are passed along to reveal as-is
+              // xxx it might be just better to copy the whole complete_config instead
+              // of selecting some names, which would allow users to transparently use
+              // all reveal's features
```
damianavila: I agree but that is some work to be done later... not in this PR.

rise/static/main.js
```diff
@@ -475,9 +480,12 @@ define([
                     condition: function() { return !!document.body.classList; } }
                 ]
               };
+              for (let setting of inherited) {
+                options[setting] = complete_config[setting];
+              }
```
damianavila: Nice!

rise/static/rise.yaml
```diff
@@ -0,0 +1,216 @@
+#
+# NOTE that the defaults defined here are essentially NOT used
+# in the code, but if only for documentation purposes it is important
+# that they reflect the defaults that are hardwired in main.js
+#
+Type: Jupyter Notebook Extension
+Compatibility: jupyter-4.3, jupyter-notebook-5.2
```
damianavila: Is this compatibility strict? I mean, if I have notebook 5.4, it will work without issues?

rise/static/rise.yaml
```diff
@@ -0,0 +1,216 @@
...
+Name: RISE - Slideshow with reveal.js
```
damianavila: Just RISE I think is enough...

rise/static/rise.yaml
```diff
@@ -0,0 +1,216 @@
...
+Main: main.js
+Description: "Run notebook as a slideshow thanks to reveal.js"
```
damianavila: I would use: `Turn your Jupyter Notebooks into a live presentation (slideshow).`

rise/static/rise.yaml
```diff
@@ -0,0 +1,216 @@
...
+Link: configurator.md
+Icon: rise.png
```
damianavila: Thanks for the icon you build. Eventually, we should come up with a nice logo, but it works for now.

rise/static/rise.yaml
```diff
@@ -0,0 +1,216 @@
...
+Parameters:
+#
+- name: rise.autolaunch
+  description: >
+     <code>autolaunch</code>: autorun slideshow upon opening if set; active only
+     on notebooks that have a 'livereveal' or 'rise' section in their
```
damianavila: Well, it will be active also with the configurator, not just the metadata... I think it is confusing as is written.

rise/static/rise.yaml
```diff
@@ -0,0 +1,216 @@
...
+- name: rise.start_slideshow_at
+  description: >
+    <code>start_slideshow_at</code>: a string that describes where to start the slideshow
+    this can be either <code>beginning</code> (start on first slide)
+    or <code>selected</code> (start on the slide that contains the currently selected cell)
+  input_type: text
+  default: selected
+# 
+# the code also has provision for setting this to 'first'
+# but that's really not useful so I leave it out of this description
```
damianavila: OK.

rise/static/rise.yaml
```diff
@@ -0,0 +1,216 @@
...
+  default: none
+- name: rise.shortcuts.toggle-skip
+  description: <code>rise.shortcuts.toggle-skip</code> shortcut to toggle the 'skip' tag on current cell
+  input_type: hotkey
+  default: none
+- name: rise.shortcuts.render-all-cells
+  description: <code>rise.shortcuts.render-all-cells</code> shortcut to put all cells in render mode
+  input_type: hotkey
+  default: none
+- name: rise.shortcuts.toggle-skip
```
damianavila: Should not be `toggle.edit-all-cells` below?

rise/static/rise.yaml
```diff
@@ -0,0 +1,216 @@
...
+
+#
+# passed to reveal as-is
+#
+- name: rise.theme
+  description: >
+    <code>theme</code>: <strong>This setting, like all the ones below,
+    is passed directly to <a href="https://github.com/hakimel/reveal.js">
+    <code>reveal.js</code> see this doc</a></strong>.
+    <br/> theme: <a href="https://github.com/hakimel/reveal.js#theming">see a list of reveal's available native themes</a>
```
damianavila: Can you add a note here about dark themes not supported? In fact, you can use a dark theme, but you need to modify some things with a custom CSS to make it work.

rise/static/rise.yaml
```diff
@@ -0,0 +1,216 @@
...
+    <code>slideNumber</code>: a string to specify if and how to number slides;
+    <a href="https://github.com/hakimel/reveal.js#slide-number" target='_'>see reveal's explanations.</a>
+  input_type: string
+  default: true
+#
+- name: rise.width
+  description: >
+    <code>width</code>: a css width attribute, like <code>1920px</code> or <code>10cm</code> or
+    <code>100%</code>.
+    <strong>Remember that you can always zoom in and out</strong>
```
damianavila: Thanks for adding this.

rise/static/rise.yaml
```diff
@@ -0,0 +1,216 @@
...
+#
+- name: rise.center
+  description: >
+    <code>center</code>: see reveal documentation
+  input_type: checkbox
+  default: true
+#
+# I prefer to keep these 2 unexposed, it's unclear how to properly enter
+# floats using the configurator UI, and it may be more harmful than helpful
+# on the long run
```
damianavila: I agree, let's keep this hidden.

rise/static/configurator.md
parmentelat: Yes, something like it; when you manage an extension in nbextensions, it shows the parameters, and below that a documentation segment
See for example autopep8 that is among the first ones to get a sense

I haven't spent any tome on that, it was just to materialize this possible place where we can publish information; it could also just be that some other  already available README could do the trick just as well; what do you think ?

rise/static/main.js
parmentelat: Oops, I feel like I missed something big here.  
When reading the code at first I thought that the 'livereveal' ConfigSection corresponded to the 'livereveal' area in the metadata;  
I have never seen nor managed a file named `livereveal.json`  
Can you please elaborate on how that is working in the current official release, so that I can adapt this code to behave in a backward-compatible way.

rise/static/main.js
parmentelat: My general impression, ever since I started using RISE, is that the 2 names livereveal and rise are confusing to many.
This is why in this release I support both names.

Once I get a better understanding of this .json thing I can reproduce here the same policy as with the notebook metadata, i.e. honor both sources of settings; in practice that would mean use settings from rise.json, then from livereveal.json if still not specified.  
But I need to understand the json thing better, of course.

rise/static/main.js
parmentelat: Yes that's exactly the idea.

rise/static/main.js
parmentelat: totally agree, it's just a note to not forget to consider that option at some later point

rise/static/rise.yaml
parmentelat: Now that I see this again, it's just what was in the file that I copied to start with, I have not tried to find the right values.
Honestly I don't know exactly what the exact impact is here. From a quick test I just made, it looks like this is purely informative; it tells users what they need to have installed to run the extension. 
 It is also possible to remove that line completely, but my feeling is that we should take this chance to spell out what the accurate dependencies are; it's just that I do not know this information for sure.

rise/static/rise.yaml
parmentelat: in next commit

rise/static/rise.yaml
parmentelat: in next commit

rise/static/rise.yaml
parmentelat: This was related to the explanation above; I will add a few more words to try and make it clearer here

rise/static/rise.yaml
parmentelat: you are right of course

rise/static/rise.yaml
parmentelat: see next commit

rise/static/configurator.md
parmentelat: Looking at this again, in particular wrt the doc2 branch that comes just next, I think we could just use the toplevel's README.md for that purpose.  
This actually leads us back to this packaging thing; I have not yet looked into that, mostly because I feel competent only for pip; but I will give this a shot shortly, I'll just let you see the conda side of things if that's all right with you.

parmentelat: After looking at setup.py, it seems to me that the pip packaging should come out OK.

Specifically there are now 6 files that should be installed in the `nbextensions/rise` area - wherever that is : 
* configurator.md
* `main.css`
* `main.js`
* `reveal.js`
* `rise.png`
* `rise.yaml`

I propose that we leave `configurator.md` as it is for now, and I will iron this out in the second PR - #347 - it will be easier there.

parmentelat: oops, correction, the PR where this will be easier to deal with is #355 because I have introduced a meta target `npm run build` that would be the right place to add a copy of the toplevel README into the static area.

regardless, I suggest we keep this issue aside for now.

rise/static/main.js
damianavila: > When reading the code at first I thought that the 'livereveal' ConfigSection corresponded to the 'livereveal' area in the metadata;

Well that's not the case ;-)

> Can you please elaborate on how that is working in the current official release, so that I can adapt this code to behave in a backward-compatible way.

If you use python to customize your config as described here: https://damianavila.github.io/RISE/customize.html#using-python

You are actually creating a `livereveal.json` file which contains the values you customized and it is read by RISE in the line I linked above.

You can try running that Python code I reference above and you will see a new `livereveal.json` file at PREFIX/etc/jupyter/nbconfig.

For instance,
```bash
 cat /Users/davila/miniconda3/envs/rise_stable/etc/jupyter/nbconfig/livereveal.json
{
  "theme": "sky",
  "transition": "zoom",
  "start_slideshow_at": "selected"
}
```

rise/static/main.js
damianavila: >is that the 2 names livereveal and rise are confusing to many.

I agree, I think it is OK to support both names for compatibility reasons but we should use rise instead of livereveal going forward.

> in practice that would mean use settings from rise.json, then from livereveal.json if still not specified.

Yep.

rise/static/main.js
damianavila: Roger that!

rise/static/main.js
damianavila: But `complete_config.is_slideshow` exists only if the notebook has some rise metadata defined. What happens with the slideshows that do not have any rise metadata defined?

rise/static/main.js
damianavila: Yep.

rise/static/rise.yaml
damianavila: > but my feeling is that we should take this chance to spell out what the accurate dependencies are; it's just that I do not know this information for sure.

You can see that we have specific RISE versions for specific notebook versions: https://damianavila.github.io/RISE/changelog.html#versions
And in setup.py we have notebook >= 5.0.0. We can keep that until a major version of the notebook is released (6.0) and then adjust properly.

rise/static/main.js
parmentelat: OK, I think I now have a more accurate mental model. 

It means in particular that the code as it is right now, as you pointed out in your first review, is not going to work properly wrt the json files. I need to find the time to fix that.

rise/static/main.js
parmentelat: Well unfortunately these notebooks won't autolaunch. I thought that would be an acceptable tradeoff.

Now that we talk about it: another approach would be to simply look for cells that are tagged as 'Slide'; that would be the criteria behind 'is_slideshow', and it would allow to more effectively classify notebooks that need to autolaunch

How does that sound ?

rise/static/rise.yaml
parmentelat: IMO the information in one `rise.yaml` should depict the actual requirement for **that** RISE version. The fact that other versions are also available is of interest too, but of second order.

I mean, if I open the configurator it means I already have a version of RISE installed and what I need to know is the compatibility versions for that one.

Also, I thought some users had recently reported issues with using 5.0, did I get that wrong ?

rise/static/main.js
damianavila: >It means in particular that the code as it is right now, as you pointed out in your first review, is not going to work properly wrt the json files

Yep.

> I need to find the time to fix that.

No problem, whenever you can!

rise/static/main.js
damianavila: > How does that sound ?

Sounds good to me! I was thinking the same when I reviewed the PR but I forgot to write it down.

parmentelat:
So for the record, and to make sure that I am dropping anything, I have 2 items on my todo list

* make sur the JSON files - bot `rise` and `livereveal` get properly loaded 
* review the policy that guesses if a notebook has slides - based on the presence of at least one slide or subslide cell

damianavila:
> I have 2 items on my todo list

Yep, those two things, I believe.

parmentelat:
This IMO should nail it; let me know how if not -- Thierry

damianavila:
> This IMO should nail it; let me know how if not -- Thierry

Thanks, I will take a look soon.

rise/static/main.js
```diff
@@ -16,79 +16,143 @@ define([
...
 
+  "use strict";
+
   /*
-   * Add customized config on top of the default options using the notebook metadata
-   * or the config-derived values
+   * load configuration
+   *
+   * 1) start from the hardwired settings (in this file)
+   * 2) add settings configured in python; these typically can be
```
damianavila: Thanks for adding this!

rise/static/main.js
```diff
@@ -16,79 +16,143 @@ define([
...
+  }
+    
+  /*
+   * this function is a heuristic that says if this notebook seems to
+   * be meant to be a slideshow.
+   * this primarily is for autolaunch, so that somebody who would
+   * enable autolaunch in her ~/.jupyter/ area would not
+   * see RISE trigger on every single notebook
+   *
+   * xxx note that this might take too long on large notebooks
```
damianavila: That's a good point, maybe we can break the loop when a slide (or subslide) is found? That should reduce the time we spend on large notebooks IF they have a slide (subslide) cell.
In any case, we can do this later, no need to change for now.

parmentelat: right; actually when a slide or subslide is found, then the function returns immediately

I was more concerned about people who would have RISE installed but who would work on large non-slides notebooks

but indeed this must not keep us from moving forward, it's an easy fix that can be made later on

parmentelat:
yes please, merge at your best convenience

PR #345, opened 2018-02-16
adding rtd link

choldgraf:
cc @damianavila


parmentelat:
the need for this change would be removed if we go for my rework of the doc in #347

choldgraf:
I'm happy with either way (though note my previous comment about gigantic PRs vs. focused PRs :-) )

damianavila:
> I'm happy with either way

OK, let's close this one because it is superseded by #347. Thanks, @choldgraf.

PR #355, opened 2018-03-01
chalkboard redeux

parmentelat:
This is a redeux for #226 

Many notes though:
* This is not quite completely working yet, but I have to brain dump this b/c I lack more time to completely finish
* this PR is built **on top of** the one on nbextensions #344, so it is much better to study after that one
* I use that configurator thing to enable/disable this new extension, and by default right now it is turned off; this way we can integrate even if not working perfectly
* In the mix, I have adopted @superbobry's change about `Reveal.initialize()` vs `Reveal.configure()` because, as he pointed out, dependencies are not applied when using the latter. And this btw has an impact on loading the notes plugin as well
* I have not found a way to install the chalkboard through `npm install`; this needs to be double checked as I am far from an expert here. So I ship this as part of our our repo - I refrained from using git subtrees here - but in the new `reveal-plugins` subdir, and not in `rise/static/reveal.js` which is an area automatically built. 
* The new aggregate target `npm run build` is  convenience that redoes everything (except `npm install`, and maybe I was wrong here)

The remaining issues are documented in the last commit; I have not yet reached the point where I could apply superbobry's cool idea of putting more stuff in css from js, will probably be needed later on to address the overlapping issues.
But I am getting too far apart and am running out of time, so I feel it's best to make a stop here.


damianavila:
This is super @parmentelat! Thanks for doing it. Will try to provide feedback in the next few days.

parmentelat:
One last note, I forgot to mention that the original PR by @superbobry introduced more separation of styling from js into css, that I liked, but have had no time to reproduce here, and but would still be worth doing IMO

damianavila:
Let's finish #344 before doing a review of this one, so I do not review the same thing twice (also difficult to see the real changes because of the noise from the other PR).
But one general thing we should do is to make the chalkboard plugin and npm package and get is from npm as we do with `reveal.js`.

It should be straightforward to make the npm package. It should be just a matter to add the proper package.json and you are done. Let me know if you need help with this. After that we can download the chalkboard from npm and we do not need to ship additional code.

Btw, there are really interesting plugins out there... I did not have time to check them but some of them are really nice, ie the menubar plugin...

parmentelat:
Agreed, I opened this issue with the original author a few days back
https://github.com/rajgoel/reveal.js-plugins/issues/46

feel free to elaborate on this issue directly, it would be more optimal if done natively at the source :)

parmentelat:
In addition, going back to the "let's put this in standby" arguments, I wanted to add this

Right now the code kind of works but it's supoptimal, in particular it behaves very poorly with input keyboard events that are defined both in the chalkboard and in the jupyter workd 

 it just occurred to me that one way to go about that problem would maybe to have these keyboard events managed by jupyter too. just an idea for now, but I'll try to give this a shot if/when I have a chance

damianavila:
> feel free to elaborate on this issue directly, it would be more optimal if done natively at the source :)

Thanks, will try to put some comments there if we don't see actions.

> it just occurred to me that one way to go about that problem would maybe to have these keyboard events managed by jupyter too. just an idea for now, but I'll try to give this a shot if/when I have a chance

Sounds like a good idea.

parmentelat:
as requested: this PR should now only show the relevant changes; hopefully at least ;)

parmentelat:
As a summary / status on this one, two topics are still unsettled

* I don't like the way keyboard events are managed in the current version, I need to try the other approach of having them all managed by the jupyter event loop

* also, the way the  chalkboard plugin code is currently distributed can be largely improved. As a reminder, for now I have duplicated this code in this repo under `reveal-plugins` but that badly sucks.
Ideally `npm install` would be the easiest by far - I just pinged the other repo about that discussion; or, if not available, with a git subtree, if we can agree on that approach.

parmentelat:
I just submitted a PR with the original chalkboard plugin

https://github.com/rajgoel/reveal.js-plugins/pull/50

that aims at publishing the chalkboard plugin under npm, and I published a test version that is available here

https://www.npmjs.com/package/reveal.js-chalkboard

If the original author is willing to follow up on that, it would mean that I can completely review/simplify the way this PR installs the plugin

parmentelat:
I have rebased my branch so as to get rid of the addition of the plugin code

keyboard event management remains as the one issue that is left

damianavila:
> I have rebased my branch so as to get rid of the addition of the plugin code

Amazing! Thanks!!

> keyboard events management remains as the one issue that is left

Do you want to work on that before I look into review this one?

parmentelat:
> > keyboard events management remains as the one issue that is left

> Do you want to work on that before I look into review this one?

well, ideally yes, but honestly my priorities about this one PR have dropped a little; so maybe it's best if you take a look at what we have right now already.

damianavila:
> so maybe it's best if you take a look at what we have right now already.

That's ok, I will try to review it soon. We can leave the keyboard thing as an enhancement for later, if it is necessary.

package.json
```diff
@@ -19,9 +20,10 @@
     "url": "https://github.com/damianavila/RISE.git"
   },
   "scripts": {
+    "build": "for target in build-css build-reveal reset-reveal; do npm run $target; done",
     "build-css": "npm run less",
-    "build-reveal": "cp -r ./node_modules/reveal.js rise/static/reveal.js",
-    "reset-reveal": "sed -i.upstream '11 s_^_\/*_' rise/static/reveal.js/css/reveal.css",
+    "build-reveal": "for dep in reveal.js reveal.js-chalkboard; do cp -r ./node_modules/$dep rise/static/$dep; done",
+    "reset-reveal": "sed -i.upstream '11 s_^_/*_' rise/static/reveal.js/css/reveal.css",
```
damianavila: There is a missing backslash on this line, are you testing this with OSX, right?
I would prefer to keep the backslash because it is harmless for OSX and I have tested successfully on linux, in the past...

damianavila: Eventually, we can use the `reveal_rise` package I created for the upcoming jupyterlab_rise, ref https://github.com/damianavila/RISE/pull/381#issue-186469852 to provide the proper CSS and get rid of this hack...

rise/static/main.js
```diff
@@ -519,21 +522,48 @@ define([
...
               }
 
-              Reveal.configure(options);
+              ////////// set up chalkboard if configured
+              let enable_chalkboard = complete_config.enable_chalkboard;
+              if (enable_chalkboard) {
+                options.dependencies.push({ src: require.toUrl('./reveal.js/plugin/chalkboard/chalkboard.js'), async: true });
+                // xxx need to explore the option of registering jupyter actions 
+                // and have jupyter handle the keyboard entirely instead of this approach
+                // could hopefully avoid conflicting behaviours in case of overlaps
```
damianavila: makes sense for later...

rise/static/main.js
```diff
@@ -519,21 +522,48 @@ define([
...
+                           63:  KeysMessager,                                        // '?' show our help
+                           61:  function() { RevealChalkboard.reset() },             // '=' reset chalkboard data on current slide
+                           45:  function() { RevealChalkboard.clear() },             // '-' clear full size chalkboard
+                           67:  function() { RevealChalkboard.toggleChalkboard() },  // 'C' toggle full size chalkboard
+                           99:  function() { RevealChalkboard.toggleChalkboard() },  // 'c' toggle full size chalkboard
+                           78:  function() { RevealChalkboard.toggleNotesCanvas() }, // 'n' toggle notes (slide-local) 
+                           110: function() { RevealChalkboard.toggleNotesCanvas() }, // 'N' toggle notes (slide-local)
+                           68:  function() { RevealChalkboard.download() },          // 'D' download recorded chalkboard drawing
+                           100: function() { RevealChalkboard.download() },          // 'd' download recorded chalkboard drawing
+                           // when 'd' is pressed
```
damianavila: I do not understand this comment...

rise/static/main.js
```diff
@@ -519,21 +522,48 @@ define([
...
+                         });
+              }
+
+
+              Reveal.initialize(options);
```
damianavila: Can you remind me why you are using `initialize`here instead of `configure`?

rise/static/main.js
```diff
@@ -663,32 +705,18 @@ define([
     var help_button = $('<i/>')
         .attr('id','help_b')
         .attr('title','Reveal Shortcuts Help')
-        .addClass('fa-question fa-4x fa')
+        .addClass('fa-question fa-3x fa')
```
damianavila: Why are you changing the size here? Smaller things fit better with the chalkboard stuff?

src/less/main.less
```diff
@@ -209,3 +209,44 @@
...
+}
+
+#help_b {
+  left: 0.6em ;
+  bottom: 0.8em;
+}
+
+/* use !important to override stuff that
+   would normally be configured through
+   chalkboard options */
```
damianavila: And why you can't configure them using the chalkboard options? Because you use a set of config options by default and there is not an easy way to access the chalkboard config mechanism?

damianavila:
@parmentelat, I left some comments.
If you can also resolve the merge conflicts (merge or rebase on master), that would be great.

rise/static/main.js
parmentelat: This is a point that the original author had pointed out, so in his own words:

> `Reveal.configure` does not load dependencies

rise/static/main.js
parmentelat: IIRC, yes that's the spirit, there are new buttons related to chalkboard, so this looked better, but it's clearly a matter of taste, feel free to undo of course

src/less/main.less
parmentelat: again it's been a long time so I'm not 100% positive here
IIRC since we load the chalkboard extension as-is, it comes with settings in its own css, but we want to override these, hence `!important`

parmentelat:
merge has been done; see also my answers to your comments
generally speaking, this in my view is in standby mode for now
I'd like to urge you to release 5.3.0 before pursuing this one

damianavila:
> I'd like to urge you to release 5.3.0 before pursuing this one

That's the plan, but I felt I needed to provide you with feedback on this one after a long time without review from me.

damianavila:
@parmentelat I did some updates and PR'ed against your branch: https://github.com/parmentelat/RISE/pull/1. If you can review my changes and test it, that would be awesome. Thanks!!

damianavila:
Closing in favor of #408. 
@parmentelat, thanks for the work you have done on this PR. I just submitted #408 with some fixes to be able to merge into master soon.

Issue #371, opened 2018-04-27
mixup between `note` and `notes` 


parmentelat:
While testing notebook-5.5 for #367, I came across another issue but in our code this time.

I seem to have mixed up on several occasions between `notes` and `note`;  
as a result there are references to an action named`RISE:toggle-note` which should be `RISE:toggle-notes`.

Trying to bind a keyboard shortcut to that inexisting action of course won't work - even with notebook-5.5

damianavila:
Thanks for the catch! Will look into the PR soon.

Btw, Github did not reply me back about your permissions in the repo. Do you currently have access to the Settings page?

parmentelat:
nope, still no settings tab for me

no big deal though, as far as I am concerned

damianavila:
> nope, still no settings tab for me

Grrr...

> no big deal though, as far as I am concerned

OK.

damianavila:
But, do you have merge permissions?

parmentelat:
> On 1 May 2018, at 16:11, Damian Avila <notifications@github.com> wrote:
> 
> But, do you have merge permissions?
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

well I have not had a chance to try yet; but I was able to tweak some issues, so I guess I have some write permissions.

so, when you say ‘looks good to me’ do you expect me to merge stuff directly ?

damianavila:
> so, when you say ‘looks good to me’ do you expect me to merge stuff directly ?

Yep, that's a correct interpretation!
If you have at least one positive review on the PR (and my LGTM accounts for that), feel free to merge it.

Issue #358, opened 2018-03-08
create shortcut to go to the configurator


parmentelat:
This again is pending a few PRs, it would maybe help to expose a Jupyter action that brings a user directly to the configurator page for RISE

My understanding is that it's just a matter of redirecting to 
`/nbextensions/?nbextension=rise/main`

damianavila:
Added by #358. Closing now.

PR #376, opened 2018-04-29
patch to address #358 - (together with a micro-change that belongs in #371)

parmentelat:
* define action RISE:rise-nbconfigurator
* bind it by default to shift-c
* corresponding addition in rise.yaml
* as well as in customization doc
* also contains one remaining change in the doc about toggle-note vs toggle-notes

### DIFF
--- a/rise/static/main.js
+++ b/rise/static/main.js
@@ -72,6 +72,8 @@ define([
         'toggle-slide': 'shift-i',
         'toggle-subslide': 'shift-u',
         'toggle-fragment': 'shift-f',
+        // this can be helpful
+        'rise-nbconfigurator': 'shift-c',
         // unassigned by default
         'toggle-note': '',
         'toggle-skip': '',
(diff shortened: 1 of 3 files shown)


rise/static/main.js
```diff
@@ -931,10 +933,24 @@ define([
...
+    actions.register({ help: 'open the nbconfigurator page for RISE',
+                       handler: nbconfigurator},
+                     "rise-nbconfigurator", "RISE");
+
     // mostly for debug / information
-    actions.register({ help   : 'show RISE config in console',
+    actions.register({ help   : 'output RISE configuration in console, for debugging mostly',
                        handler: showConfig},
-                     "show-config", "RISE");
+                     "rise-dump-config", "RISE");
```
damianavila: `showConfig` is commented in the [source code](https://github.com/parmentelat/RISE/blob/master/rise/static/main.js#L1033), maybe we can make it an option through config as well? Or do you want to keep it hidden?

rise/static/main.js
```diff
@@ -931,10 +933,24 @@ define([
                      },
                      "edit-all-cells", "RISE");
 
+    // because the `Edit Keyboard Shortcuts` utility does not mention the
+    // actions prefix (i.e. 'RISE' in our case), we choose to make these two
+    // action names start with `rise-` even if it's a bit redundant.
+
+    // define an action that goes to the nbconfigurator page for rise
```
damianavila: Good idea!

rise/static/main.js
parmentelat: yeah, show-config really is something that end users won't care about at all, it just dumps the config struct in the javascript console..

Issue #357, opened 2018-03-08
readme in configurator


parmentelat:
Once the dust from #344 #347 #355 settles, it might make sense to revisit the doc displayed in the configurator, which for now is a dummy placeholder called `configurator.md`, but that could be fruitfully replaced with the main README.

parmentelat:
referring to #377,

when we do the packaging for 5.3.0, we simply need to package the toplevel README.md file together with main.js and main.css and the rest of the static files

in that case it will be enough to 
* replace in `rise.yaml`  the reference to `configurator.md`  with  `README.md`
* and then remove `configurator.md` altogether from the git repo

damianavila:
PR was merged, let's close this one.

PR #379, opened 2018-05-02
tentative review of setup.py

parmentelat:
for using only setuptools, and simplifying as much as possible

at the same time this would take care of #357

Damian please let me know what you think about going in this direction  
I feel like I may be missing something here, so I don't want to spend more time until you give me the green light, in case this is a dead end

### DIFF
--- a/examples/font-sizes-untampered.ipynb
+++ b/examples/font-sizes-untampered.ipynb
@@ -1,0 +1,1 @@
+font-sizes.ipynb
--- a/examples/font-sizes.css
+++ b/examples/font-sizes.css
@@ -1,0 +1,23 @@
+/* an example to illustrate how to tweak font sizes */
+
+/* to apply a setting only when under RISE, 
+   be aware that the <body> tag has class .rise-enabled */
+
+
+/* ---------- markdown cells
+ */
+body.rise-enabled div.inner_cell>div.text_cell_render.rendered_html {
+    font-size: 80%;
+}
+
+
+/* ---------- code blocks inside markdown
+   i.e. within ``` lines, or 4-space indented
+ */
+div.inner_cell>div.text_cell_render.rendered_html>pre {
+    margin: 0px;
+}
+
+div.inner_cell>div.text_cell_render.rendered_html>pre>code {
+    font-size: 70%;
+}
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -1,63 +1,75 @@
...
+    if target_dir == 'share/jupyter/nbextensions/rise':
+        files.append('README.md')
     data_files.append((target_dir, files))
+
 data_files.append(('etc/jupyter/nbconfig/notebook.d', ['rise.json']))
 
-version_ns = {}
-with open(pjoin(here, name, '_version.py')) as f:
-    exec(f.read(), {}, version_ns)
+data_files
```
damianavila: What is the purpose of this line?

setup.py
```diff
@@ -1,63 +1,75 @@
...
+
+INSTALL_REQUIRES = [
+    'notebook>=5.5',
+]
 
 with open('./README.md') as readme:
     README = readme.read()
 
+for d in data_files:
```
damianavila: debugging stuff? If you really want to show the content of the data_files object, I would put this piece after line 51, I think...

damianavila:
@parmentelat, I left some minor comments, I think this is going in the right direction. Do you want to push further on this? Otherwise, I will merge it after your answers/clarifications (and after testing the whole installation with this change).

parmentelat:
oops it looks like I screwed up here, because I used the name 'master' in my own repo, which I thought was to be in line with yours

in any case closing this was unintentional, let me fix this in my repo first

parmentelat:
look Damian, I am terribly sorry; I've been able to resurrect the commit in question, but no matter how hard I tried, I have not yet been able to retrieve your comments on this PR

I am going to have to create a new PR with the exact same contents; I'll keep on looking for a way to resurrect your comments but am not very optimistic

deepest apologies again for that screw up;
unfortunately I had not taken the time to push this under a new name, and had been using 'master'
in the meanwhile I totally forgot about that, and my mental model was that 'master' on my own repo was safe to force-push; it was not, obviously :(

damianavila:
> I am going to have to create a new PR with the exact same contents; I'll keep on looking for a way to resurrect your comments but am not very optimistic

No problem.

> deepest apologies again for that screw up;

No need to apologize.

PR #387, opened 2018-05-22
tentative review of setup.py

parmentelat:
re-creating #379

for using only setuptools, and simplifying as much as possible

at the same time this would take care of #357

Damian please let me know what you think about going in this direction
I feel like I may be missing something here, so I don't want to spend more time until you give me the green light, in case this is a dead end


parmentelat:
Your initial review was on the first of these commits

While I was at it, I added another commit on top of what you'd already reviewed; this 2nd commit should be reasonably straightforward

Again let me know how this feels like; again it's not been tested at all; my take on this is, with the requirement for a new packaging (with nbconfigurator and all), we should as well do it within this more modern framework, as `distutils` so definitely sucks ;)

parmentelat:
giving this a deeper try:

I created a placeholder under test.pypi.org and uploaded there using my experimental `setup.py`, which allowed me to test pip3-based installs from a fresh VM 

I am facing an issue though, that has to do with setuptools's policy to not encourage, or maybe even not support as far as I understand it, to install stuff in places other than under your own package directory

FYI, I've been precisely migrating to setuptools for other projects of mine; as an example I used to ship stuff that installed directly under `/etc/`, but this did not seem possible with setuptools; which felt frustrating at first, but actually is a good idea since '/etc' for example is not something that makes sense on Windows..

Regardless, coming back to RISE, I came to realize that I do not quite understand the requirements here; for example there are mentions of paths in `share/jupyter/nbextensions/rise` and `etc/jupyter/nbconfig/notebook.d`, but I do not fully understand what these paths are relative to at install-time.

Of course it feels like this is all about installing a copy of 'static' somewhere where jupyter can find it, but I am not enough comfortable with jupyter's extensions to get this right

Can you shed some light on this, or point me to the place in jupyter's documentation where I can find the necessary details to complete this move ?

----
I take this chance to dump a few useful commands when doing these tests

* uploading on test.pypi.org

  * there's a need to artificially bump the version number each time you upload; I've already burned 5.3.0 and 5.3.1 (but this is only relative to test.pypi.org, so we will still be able to issue a proper5.3.0 on the official pypi.org)
  * there is a need to configure your test.pypi.org account once and for all as described here <https://packaging.python.org/guides/using-testpypi/#setting-up-testpypi-in-pypirc>
  * then `python3 setup.py sdist upload -r testpypi`
  * if you wanted to try this out, I'll probably need to grant you write access on this project..

* installing on a fresh VM

  * start with installing the stable release to pull all the dependencies:`pip3 install rise`
  * then to try the fake one: `pip3 install -U --index-url https://test.pypi.org/simple rise`

parmentelat:
OK, I found all the answers I think, and so this third commit should mostly nail it, I was even able to test it

So please review asap ! 

It would be best IMHO to expedite this matter while it's hot, as the overhead for warming up is particularly expensive, for me at least :)

parmentelat:
cool; I guess there's nothing left between us and a 5.3.0 release then :)

PR #388, opened 2018-05-26
Minor fixes preparing the next release.

damianavila:
* Docs
* Versions

@parmentelat anything else you think we need to add here?

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -15,15 +15,15 @@ build:
 requirements:
   build:
     - python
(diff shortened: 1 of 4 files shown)


parmentelat:
Sent from my iPhone

> On 27 May 2018, at 00:42, Damian Avila <notifications@github.com> wrote:
> 
> Docs
> Versions
> @parmentelat anything else you think we need to add here?
> 

Not off the top of my head
Just remember that the readthedocs/github peering might need a review once you release

Issue #393, opened 2018-05-30
version id management


parmentelat:
the fact that the version id needs to be manually changed in a string of files - 4 if the develop page is right - is quite a pain

in addition, I would argue that this version id should be altered in develop branches; for example now that we just issued 5.3.0, the version id for the master should be something like `5.3.1.devel` or similar, anything that's **not** 5.3.0

adopting such a workflow would mean having to change this version id more often, but having to change it in 4 different locations is both tedious and error-prone

I am unsure what to propose specifically, but wanted to start a discussion on this matter; there might be standard approaches to that ? otherwise I can come up with a `npm run version` utility, or something similar.

damianavila:
> the fact that the version id needs to be manually changed in a string of files - 4 if the develop page is right - is quite a pain

I agree.

> in addition, I would argue that this version id should be altered in develop branches; for example now that we just issued 5.3.0, the version id for the master should be something like 5.3.1.devel or similar, anything that's not 5.3.0

I agree as well.

> adopting such a workflow would mean having to change this version id more often, but having to change it in 4 different locations is both tedious and error-prone

Yep.

> I am unsure what to propose specifically, but wanted to start a discussion on this matter; there might be standard approaches to that ? otherwise I can come up with a npm run version utility, or something similar.

We should look for a unified approach, something like this: https://github.com/msarahan/state-adjacency-graphs/blob/master/conda.recipe/meta.yaml#L20-L23, but also including package.json and docs reference... maybe a script automating all these changes?

damianavila:
Closed by #399, removal of the conda recipe will be done #405.

PR #399, opened 2018-07-02
fixes #392, and provides a first approximation for issue #393

### LINKED ISSUE #392: stable documentation does not show version
as a result of #307 readthedocs now points at the latest stable version

this is cool; however when landing on that page it is no longer clear which version the documentation is about - or I am missinf it - so it would be nice if the front page somehow included the version id

parmentelat:
reducing the number of places where a hard-wired version identification string needs to be defined to 2:
* package.json
* conda.recipe/meta.yaml

the latter might need some additional work to be removed, but I am not familiar enough with conda to take that into account

one notable side-effect is the need to package `package.json` right under the `rise` module so that it can be located with `pkg_resources` (which comes with setuoptools);
hence the symlink `rise/package.json` that points one step up

please also note that `version_info` - formerly unused as far as I can tell - is now returned as a <Version> object as the result of `pkg_resources.parse_version`

---
this release defines version as being `5.4.dev0` which, my understanding is, is what PEP440 foresees for a development version


parmentelat:
Unfortunately I have not been able to check on readthedocs how that is going

The reason is that the branch name that I used 'issue393' is not taken into account over there for some reason, probably some additional configuration is needed...

Also note that I have not even tried to deal with the conda recipe as I am totally unfamiliar with conda, if some else can provide directions I'd be happy to wrap this up

imagejan:
> * conda.recipe/meta.yaml

Maybe this paragraph from the [conda package docs](https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#package-section) helps here:

> NOTE: Post-build versioning: In some cases, you may not know the version, build number or build string of the package until after it is built. In these cases, you can perform [Templating with Jinja](https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#jinja-templates) or utilize [Git environment variables](https://conda.io/docs/user-guide/tasks/build-packages/environment-variables.html#git-env) and [Inherited environment variables](https://conda.io/docs/user-guide/tasks/build-packages/environment-variables.html#inherited-env-vars).

parmentelat:
Digging this a little deeper, there's this link here
https://stackoverflow.com/questions/38919840/get-package-version-for-conda-meta-yaml-from-source-file

that discusses this general question in more details; in particular with the present tip of `issue393` one can indeed do

`python setup.py --version` 

to programmatically retrieve the version number set in `package.json` 

So, it looks like one path forward would be to

* remove the definition of `version` in `meta.yaml`
* modify `release.md` to instruct that the conda stage can be done with e.g.

`RISE_RELEASE=1 VERSION=$(python3 setup.py --version) conda build conda.recipe --python=3.6`

Alternatively we could also script that, for example as a `npm run conda` target or something similar.

Thoughts ?

parmentelat:
@damianavila would you have a few minutes to tell me your gut feeling here ?
my understanding is, you are the primary/only person who uses the idioms to rebuild a conda release, so your input would be most helpful to figure a way forward on this one
-- Thierry

damianavila:
Hi @parmentelat!

> my understanding is, you are the primary/only person who uses the idioms to rebuild a conda release, so your input would be most helpful to figure a way forward on this one

Going forward we will use conda-forge to build our conda packages: https://github.com/conda-forge/rise-feedstock

We can probably remove the recipe from this repo. And with that removal, the remaining things in this PR should be zero ;-)

Btw, I will try to review this PR in the next few days (alongside with the other ones open).

I want to have a release with everything in PRs except the jupyterlab extension one by the end of the month.

damianavila:
This LGTM, let's merge it and iterate if it is needed. Opening a new issue: https://github.com/damianavila/RISE/issues/405, to remove the conda recipe from this repo.

PR #412, opened 2018-08-20
Speaker notes working now

### LINKED ISSUE #174: Speaker view does not display slides
Once in a reveal slideshow, pressing the `S` key opens the reveal.js speaker view, however the windows for current and upcoming slides display the normal jupyter notebook page.

Workaround:
I've found that it is possible within the tiny slide views to open up slideshow mode, and advance the "upcoming slide" window by 1 slide. This works, but it is quite annoying to do each time.

damianavila:
This PR enable speaker notes in RISE.
We just needed to customize the notes plugin (published as an npm package) to look properly for the `notes.js` file and to configure properly the open keyboard event (to avoid overriding jupyter `s` event) and we also needed to configure it properly in main.js (as the reveal.js docs suggested).
Finally we modified the autoLaunch function to launch the presentation view when you are in the speaker notes view.

Closes #174

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "less-plugin-clean-css": "~1.5.0",
     "reveal.js": "~3.5.0",
     "reveal.js-chalkboard": "~1.0.0",
+    "notes_rise": "~1.0.0",
     "watch": "~0.16.0"
   },
   "license": "BSD-3-Clause",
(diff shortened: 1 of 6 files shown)


rvoleti89:
This is great news! I have tried launching speaker notes in RISE 5.4.1 (installed with conda) by pressing the usual "S" key but nothing seems to be happening. Is there a configuration option that needs to be enabled somewhere? Did not see anything in the official documentation.

damianavila:
Sorry about this, I need to update the docs. Currently, it is not customizable yet, but since "S" was conflicting with other notebook actions, I used "T" to start the notes

rvoleti89:
Makes a lot of sense, thank you! Works perfectly.

PR #439, opened 2019-01-13
Fix scrollbar below cells with inline code

### LINKED ISSUE #430: Markdown cells with verbatim blocks get horizontal scroll bars.
I am loving using RISE in my undergrad python classes, but I have a problem:

Any markdown cell that has a verbatim block gets horizontal scroll bars in the slideshow. 

For example a cell:

```
This is a test

This is not a block
```

renders fine, but

```
This is a test:

    This is a block
```

Does get a horizontal scroll bar. The funny thing is, it doesn't matter how wide the broswer window is, the bar is always there. I can make the window half my screen, all my screen or zoom right out, but the bar is still there. 

I've tried both indented blocks and github style tripple backtick blocks, with the same result each time.

CaptainProton42:
This fixes #430 an issue where horizontal scrollbars would appear below cells containing inline code by adding a small padding to every .text_cell_render.

### DIFF
--- a/src/less/main.less
+++ b/src/less/main.less
@@ -80,9 +80,9 @@
     }
     
     /* This fixes an issue where the scrollbar would appear below cells containing inline code. */
-	  .text_cell_render {
-	   padding-right: 5%;
-	  }
+    .text_cell_render {
+      padding-right: 5%;
+    }
     
     .reveal-skip, .reveal-notes { display: none; }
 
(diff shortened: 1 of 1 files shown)


damianavila:
Thanks for the PR @CaptainProton42!!!
I will review and test as soon as possible.

damianavila:
@CaptainProton42 I found that setting up overflow-x hidden in this case is more reliable and without interactions with other things. So, I created a PR for that: https://github.com/damianavila/RISE/pull/459.
I will close this one but I want to truly thank you for your PR and your help with this issue!!

Issue #452, opened 2019-03-25
is damianavila82 still relevant ?


parmentelat:
I think I understood that damianavila82 was no longer a relevant channel, but I still find occurrences of that name: 

```bash
$ git ls-files | xargs grep damianavila82
doc/dev/release.md:**Step 8.** Upload conda packages to anaconda.org/damianavila82
doc/dev/release.md:    anaconda upload -u damianavila82 conda_dist/linux-32/*
doc/dev/release.md:    anaconda upload -u damianavila82 conda_dist/linux-64/*
doc/dev/release.md:    anaconda upload -u damianavila82 conda_dist/osx-64/*
doc/dev/release.md:    anaconda upload -u damianavila82 conda_dist/win-32/*
doc/dev/release.md:    anaconda upload -u damianavila82 conda_dist/win-64/*
doc/installation.rst: conda install -c damianavila82 rise
```

please instruct how to properly expose this matter in the docs

damianavila:
We should remove any references because we are using conda-forge for a long time now.

parmentelat:
could you please elaborate on how the doc should be updated ? replace with another name ? remove step 8 altogether ?

damianavila:
Sure...
* Step 6 and 8 should be removed because that is done (in a better way) by conda-forge.
* Keep 7 because that will upload the python package to pypi.
* Create a new step at the end where you detailed the changes you need to do in the conda-forge repo (that repo contains the conda recipe to build the conda packages).
  * First read this section: https://github.com/conda-forge/rise-feedstock#updating-rise-feedstock
  * You need to update the version number here: https://github.com/conda-forge/rise-feedstock/blob/master/recipe/meta.yaml#L1
  * You need to update the sha number here: https://github.com/conda-forge/rise-feedstock/blob/master/recipe/meta.yaml#L9
  * (Optional) You need to update any dependencies if you have new ones or remove old ones.
  * (Optional) You could update the recipe, for instance, eventually, we will get rid of the post-link steps (see, https://github.com/damianavila/RISE/pull/444).

parmentelat:
all right, will create a PR in this spirit so you can review

damianavila:
PR was merged, let's close this one.

PR #457, opened 2019-04-18
hopefully this would take care of #452

parmentelat:
please review while you are at it

no brainer I guess, it's better than before and won't break the code :-)


damianavila:
Lovely, let's get this merged!

PR #459, opened 2019-04-21
Fix x-scrolling bar in markdown cells containing code

damianavila:
This is superseding #439.


damianavila:
Tested successfully, merging now.

PR #463, opened 2019-04-24
New versioning + other stuff

### LINKED ISSUE #409: npm is not happy about the python dev version.
The current version in master `5.4.dev0` is compatible with Python PEP 440 but incompatible with the semantic version scheme from npm.
The immediate solution is changing that version to 5.4.0, but eventually, we should explore some way to have a compatible version for the dev time on master.

cc @parmentelat

damianavila:
* Enhance docs
* New versioning scheme so we can use dev version from master and fix weird 0.0.0 version number I have seen in docs and some installations
* Bring new notes_rise 1.0.1 plugin
* Fix old email
 
Fixes #409.


damianavila:
cc @parmentelat because I changed some versioning stuff you did before, I think it is more reliable right now. Merging now so can finish stuff and, finally, release 5.4.2.

damianavila:
If you have 10 min, can you create a local package and test installation with pip from that local package:

To get the package:
```
git clean -fdx
npm install
npm run build
npm run build-css
python setup.py sdist
```

To install from that package (now under dist), first create a clean environment and then:
```
pip install dist/rise-5.4.2.dev0.tar.gz
```
Btw, docs from master live at https://rise.readthedocs.io/en/latest/index.html.

I have also seen some padding issue in the cells... the space at the bottom of the text (in code cells, maybe in md cells as well) seems somehow short. Resizing the window seems to fix it partially, do you see the same?

damianavila:
> I have also seen some padding issue in the cells... the space at the bottom of the text (in code cells, maybe in md cells as well) seems somehow short. Resizing the window seems to fix it partially, do you see the same?

OK, I think I fixed that adding some margin-bottom: https://github.com/damianavila/RISE/pull/467

Issue #462, opened 2019-04-24
CCS styling is not removed when we get into the notebook view


damianavila:
Here: https://github.com/damianavila/RISE/blob/master/rise/static/main.js#L496-L502

We are allowing to inject css stuff, and that css is applied when we enter the slideshow view, but when we go back to the notebook view, that css is still applied which could, in turn, make the notebook view look very weird depending on the css changes you introduced.
This can be actually checked with the current example/README.ipynb

cc @parmentelat, I am pretty sure you will be interested in this since you modified the example recently.

We could probably `disable` the `link` when we go out from the slide view.
Not a blocker for the release, but I found out this testing the example and I think it would be nice to fix it in the future...

parmentelat:
that makes perfect sense indeed, I have never had to disable css code before, will be interesting :)

parmentelat:
this should be fixed by now

PR #466, opened 2019-04-24
attempt at fixing #462

parmentelat:
use jquery remove() to remove these 2 <link> elements, at the same time as the reveal material is removed as well
and  in the mix, adopting a more sensible naming scheme for these 2 <link> elements

this is entirely untested

### DIFF
--- a/rise/static/main.js
+++ b/rise/static/main.js
@@ -492,21 +492,13 @@ define([
       let name_css = name.replace(".ipynb", ".css");
       // typically /files/examples/
       let prefix = `/files/${path.replace(name, '')}`;
-      // rise.css stylesheet: either insert a <link> object,
-      // or enable it if present
-      if ($('#rise-custom-css').length == 0) {
-        $('head').append(
+      // Attempt to load rise.css
+      $('head').append(
           `<link rel="stylesheet" href="${prefix}rise.css" id="rise-custom-css" />`);
-      } else {
-        $('#rise-custom-css').prop('disabled', false);
-      }
-      // same with the stylesheet named after the notebook
-      if ($('#rise-notebook-css').length == 0) {
-        $('head').append(
+      // Attempt to load css with the same path as notebook
+      $('head').append(
           `<link rel="stylesheet" href="${prefix}${name_css}" id="rise-notebook-css" />`);
-        } else {
-          $('#rise-notebook-css').prop('disabled', false);
-        }
+
     }
 
     function toggleAllRiseButtons() {
@@ -811,8 +803,8 @@ define([
 
     $('#theme').remove();
     $('#revealcss').remove();
-    $('#rise-custom-css').prop('disabled', true);
-    $('#rise-notebook-css').prop('disabled', true);
+    $('#rise-custom-css').remove();
+    $('#rise-notebook-css').remove();
 
     $('.backgrounds').hide();
     $('.progress').hide();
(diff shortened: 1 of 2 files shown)


damianavila:
I have read that `remove` - ing the link prevents to load it again, this is why I suggested to `disable` it.

parmentelat:
ah ! my bad if that's the case  
I have to admit that I haven't thought that through  
I naively - and too quickly - concluded that if removing the reveal css was effective, then it would work in our case too, but now I understand how it could not be true
I di search for ways to `disable` a `<link>` tag at first, but could not spot it right away; if you could please share a link towards the relevant documentation I'd be interested to fix that

damianavila:
Some interesting resources: https://stackoverflow.com/questions/3182840/removing-or-replacing-a-stylesheet-a-link-with-javascript-jquery

Btw, can you take a look into this comment? https://github.com/damianavila/RISE/pull/463#issuecomment-486055243

I am particularly worried about the last part:

> I have also seen some padding issue in the cells... the space at the bottom of the text (in code cells, maybe in md cells as well) seems somehow short. Resizing the window seems to fix it partially, do you see the same?

I would love to see if you are seeing the same...

rise/static/main.js
```diff
@@ -492,13 +492,21 @@ define([
       let name_css = name.replace(".ipynb", ".css");
       // typically /files/examples/
       let prefix = `/files/${path.replace(name, '')}`;
-      // Attempt to load rise.css
-      $('head').append(
+      // rise.css stylesheet: either insert a <link> object,
+      // or enable it if present
+      if ($('#rise-custom-css').length == 0) {
```
damianavila: What about using `$("head").is("#rise-custom-css")` (untested)?

rise/static/main.js
```diff
@@ -492,13 +492,21 @@ define([
...
+        $('head').append(
           `<link rel="stylesheet" href="${prefix}rise.css" id="rise-custom-css" />`);
-      // Attempt to load css with the same path as notebook
-      $('head').append(
-          `<link rel="stylesheet" href="${prefix}${name_css}" id="notebook-custom-css" />`);
-
+      } else {
+        $('#rise-custom-css').prop('disabled', false);
```
damianavila: Should not be `enabled` here?

rise/static/main.js
```diff
@@ -492,13 +492,21 @@ define([
...
+      }
+      // same with the stylesheet named after the notebook
+      if ($('#rise-notebook-css').length == 0) {
```
damianavila: Same comments for this section.

parmentelat:
I have been able to spend a little more time on this matter  
actually your linked SO article is out-dated, it was opened in 2010 and at that time there were still some crappy browsers around

one of the contributors took the time to update the matter in 2017:
> stuff like $('link[title="mystyle"]').remove() works now, it was an issue with old browsers (when this question was asked years ago) – Ivan Castellanos Oct 27 '17 at 20:53

so, I have moved back to my first naive solution

plus, I have added a setting in `examples/rise.css` that applies to regular notebook cells;  
it thus allows to check that this feature works as expected: when quitting reveal, the effect of that setting should vanish

this works for me on macos/chrome at least

parmentelat:
just a kind ping on this one; now that 5.5 is out, it's my feeling that we could merge this one now

it'd help me greatly to minimize the number of my local branches :)

as I said earlier, the following last lines in `examples/rise.css` are there so that we can easily convince ourselves that #462 is properly fixed; just quit reveal mode in this notebook and check that the setting is no longer applying.

```
/* this is only to check that rise.css properly gets
 * ignored when quitting reveal mode */
div.text_cell_render.rendered_html {
    color: #5050b0;
}
```

Issue #438, opened 2019-01-12
Jupyter Notebook RISE distorts layout of the HTML/CSS presentation


damianavila:
Reported here: https://github.com/jupyter/notebook/issues/4296

damianavila:
@Foadsf, my guess is something about reveal.css conflicting with the CSS you have put there... you probably need to do some exploration to know what rules are conflicting and how to overwrite them successfully.

parmentelat:
There is a lot to say here.

It is important to understand that there are actually several styling sources that are at work when you run a RISE presentation: 

* first, everything runs as a notebook so all the jupyter styling - and [typically notebook.less](https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/less/notebook.less) among others - is in effect; 
* second, `reveal.js` is involved as well of course as the slide presentation engine, and this brings in the corresponding theme - see https://github.com/hakimel/reveal.js#theming - default within RISE being `simple`;

Also, regarding your remark on using classes versus ids because of the css syntax being misinterpreted as Python comments: when using RISE there are other and safer means to inject CSS into the mix, as described in this link https://rise.readthedocs.io/en/docs_hot_fixes/customize.html#adding-custom-css

Issue #435, opened 2018-12-14
Difficulty with themes


stoffprof:
When I use certain themes, only some elements of the theme are actually being used.

For example, if I use the sky theme, all colors and fonts seem to work as expected. But if I switch to moon, the background color is white rather than the color specified in moon.

In presentation mode, when I use the Chrome inspection tool, the styles pane shows:
![capture](https://user-images.githubusercontent.com/12434323/50032789-da3df800-ffc3-11e8-82ea-07f4d27678e1.png)

And at the top it shows:
![capture2](https://user-images.githubusercontent.com/12434323/50032824-fcd01100-ffc3-11e8-8854-aa57ca8182fc.PNG)

Not sure if I'm interpreting this correctly, but it looks like the background color in moon.css is being overwritten by main.css. Am I doing something wrong? I'm just specifying the theme in a rise.json file in .jupyter\nbconfig that says:
```
{
  "theme": "moon"
  "transition": "concave",
  "start_slideshow_at": "selected",
}
```

Other font colors are not being overwritten, as can be seen by the heading color below, which does come from moon.css.

![capture3](https://user-images.githubusercontent.com/12434323/50032965-a1525300-ffc4-11e8-824f-1b907cd64f4b.PNG)

damianavila:
IIRC moon is one of the dark themes. Regrettably, dark themes need some further customization to be able to work correctly. I did not find time yet to push this forward but I would love to see new people involved in this.

stoffprof:
If you can share a little more about what the issue is, I can try to spend some time on it. At this point I don't really see where to start.

damianavila:
Essentially you need to explore the reveal CSS and how it is conflicting with the notebook CSS.
For instance, knowing how the moon theme should look like, you can see which reveal CSS rules are conflicting with the notebook ones and modify them on the fly to see if you can recover the look you pretend. I would start playing with these CSS rules in the browser JS console.

albz:
Hi, I got here since I am trying to use RISE and I would need it in a *dark mode* for my presentation.
I am actually trying to use a dark theme but I cannot find any specific easy documentation where it shows how to get to use a different theme. Can you please link me to any aspects where a tutorial like page can drive me thought the process?

damianavila:
You can use different themes as instructed here: https://rise.readthedocs.io/en/stable/customize.html#choosing-a-theme

I should mention, we are currently working into improving themes support: https://github.com/damianavila/RISE/pull/475

[...]

PR #475, opened 2019-05-18
Reveal 3.8.0

parmentelat:
this is WIP and not working; most obvious broken feature at this point is themes, and more specifically theme backgrounds, that do not render as expected

I'm exposing this so that others can experiment with and feedback about an angle that I think may be worth fiddling with

in a nutshell: as mentioned a few times in some other issues like #435 and #438, when running a RISE slideshow we primarily enable 3 source of css styling: jupyter, RISE and reveal

the way reveal themes are implemented for example sets attributes on `body` which tends to potentially have rather devastating effects

the angle I have taken in this one-line approach is to monkey-patch the official reveal themes (as downloaded after `npm install`) so as to make their `body` css rule more specific, and to give them a chance to override stuff - for example to override settings made in `notebook.less` on the `body` tag too.

---
this at this point is not working; it does improve the situation a bit, in that when inspecting the <body> tag through web devel tools, the theme settings do show up as applying, but for some yet-to-be-figured reasons the background still gets rendered blank..

Any hint would be btw much appreciated about that matter, it currently beats me why background themes don't get rendered.

### DIFF
--- a/examples/rise.css
+++ b/examples/rise.css
@@ -1,9 +1,9 @@
 /* this goes with my global default setting
  * that has rise use reveal's theme sky
  */
-.reveal {
+/*.reveal {
     font-family: "Quicksand", sans-serif;
-}
+}*/
 
 .reveal h1, .reveal h2, .reveal h3, .reveal h4, .reveal h5, .reveal h6 {
     text-transform: initial;   /* sky.css says uppercase */
--- a/package.json
+++ b/package.json
@@ -30,5 +30,5 @@
     "less": "PATH=./node_modules/.bin:$PATH lessc --autoprefix src/less/main.less rise/static/main.css",
     "watch-less": "./node_modules/.bin/watch 'npm run less' src/less"
   },
-  "version": "5.5.0"
+  "version": "5.6.0-dev0"
 }
--- a/package.json
+++ b/package.json
@@ -21,8 +21,9 @@
     "url": "https://github.com/damianavila/RISE.git"
   },
   "scripts": {
-    "build": "for target in build-css build-reveal reset-reveal; do npm run $target; done",
+    "build": "for target in build-css patch-themes build-reveal reset-reveal; do npm run $target; done",
     "build-css": "npm run less",
+    "patch-themes": "sed -i.patch 's|^body {|body.notebook_app.rise-enabled {|' node_modules/reveal.js/css/theme/*.css",
     "build-reveal": "for dep in reveal.js reveal.js-chalkboard notes_rise; do cp -r ./node_modules/$dep/ rise/static/$dep/; done",
     "reset-reveal": "sed -i.upstream '11 s_^_/*_' rise/static/reveal.js/css/reveal.css",
     "clean-reveal": "for dep in reveal.js reveal.js-chalkboard notes_rise; do rm -rf rise/static/$dep; done",
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "scripts": {
     "build": "for target in build-css patch-themes build-reveal reset-reveal; do npm run $target; done",
     "build-css": "npm run less",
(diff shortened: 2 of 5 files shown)


damianavila:
> the angle I have taken in this one-line approach is to monkey-patch the official reveal themes (as downloaded after npm install) so as to make their body css rule more specific, and to give them a chance to override stuff - for example to override settings made in notebook.less on the body tag too.

I like that idea... even if it is not working at first... but I think it is a sensible path to investigate and push forward. I will try to run your branch soon to see if I can help with some more specific thoughts.

parmentelat:
hi Damian

rendering the themes properly turned out much harder than I anticipated, but I have eventually been able to do some progress on this topic

please give this a try when you get a chance

I'm rather pleased with the results in terms of theme rendering, although the approach is admittedly a little hacky; or not even a little ;) - I really don't see how to do otherwise, your feedback will be most welcome 

---
FYI: if you are interested, I had initially gone even further in terms of theme integration
https://github.com/parmentelat/RISE/tree/more-theming-from-reveal

but that turned out too intrusive (code cells ended up way too small) so I've decided to keep that off the discussion for now at least; you may want to give that a try too if you're curious; but that again is just an FYI

parmentelat:
I'm removing the `[WIP]` from this PR's title; as far as I am concerned this seems to be closing in

damianavila:
> rendering the themes properly turned out much harder than I anticipated, 

I was expecting that, actually...

> please give this a try when you get a chance

Sure!

> I'm rather pleased with the results in terms of theme rendering, although the approach is admittedly a little hacky; or not even a little ;) - I really don't see how to do otherwise, your feedback will be most welcome

I love hacky! Jaja... I actually did some hacky things in the past so let's continue with the tradition :wink:

> FYI: if you are interested, I had initially gone even further in terms of theme integration
https://github.com/parmentelat/RISE/tree/more-theming-from-reveal
> but that turned out too intrusive (code cells ended up way too small) so I've decided to keep that off the discussion for now at least; you may want to give that a try too if you're curious; but that again is just an FYI

OK, I will take a look!

damianavila:
I will do a deep review as soon as I can, but looking around on the PR, it seems you are doing the same sort of changes with patch-themes and reset-reveal. Maybe it is now time to put those things together?

Another thing... we could be shipping a "new" npm package (as I did with the notes_rise npm package) built from that script you have written (maybe even include the notes_rise customizations in that package). If we do something like that, we abstract the hacky changes from RISE itself into an upstream package (that we control) which has all the customizations collected in just one place.

parmentelat:
ok, I'm not sure that I follow completely but that sounds about right

from a completely different angle: I'd like to be able to build docker images with my version inside, so that I can test more thoroughly

I'm not quite sure how to do that; pip has the option to install from a git repo; but I'd be surprised if that worked as-is - I must admit that I have not even tried - because pip would not know to run npm install and the other build targets; or would it ? 

is there any chance I/we can publish this dev version in a way that could be simply installed from a Dockerfile ?

damianavila:
>is there any chance I/we can publish this dev version in a way that could be simply installed from a Dockerfile ?

I would say let's push on this one, merge it and then we can release a 5.6.0 beta version. In fact, we should be releasing more betas as long as we merge potentially disruptive stuff...

parmentelat:
taking up on your suggestion, I have merged this onto master and published 5.6.0.dev1 on pypi

try out with `pip install --pre rise`

damianavila:
This is great @parmentelat!! I will test it soon.

Can we make a new issue to tackle some of the things I outlined in my quick review?

> I will do a deep review as soon as I can, but looking around on the PR, it seems you are doing the same sort of changes with patch-themes and reset-reveal. Maybe it is now time to put those things together?

> Another thing... we could be shipping a "new" npm package (as I did with the notes_rise npm package) built from that script you have written (maybe even include the notes_rise customizations in that package). If we do something like that, we abstract the hacky changes from RISE itself into an upstream package (that we control) which has all the customizations collected in just one place.

Additionally, we should probably release these dev versions with conda-forge as well...

Issue #479, opened 2019-06-19
Custom css not found on binder


sbrunk:
I'm trying to publish RISE enabled runnable slides via binder, and I'm running into an issue with custom css. Seems like it tries to load rise.css from the wrong location (https://hub.gke.mybinder.org/hub/files/examples/rise.css) and gets a 404.

I'm not sure if this is really an issue with RISE or rather caused by binder/jupyterhub because its working perfectly when running a jupyter server locally.

You should be able to reproduce and see the issue by running the [binder example in this repo](https://mybinder.org/v2/gh/damianavila/RISE/master?filepath=examples%2FREADME.ipynb) and looking into the browser dev tools.

parmentelat:
hey

no need to use your repo, the one that we have on binder:  
https://rise.readthedocs.io/en/stable/resources.html#github-readthedocs-gitter-youtube  
exhibits that issue all right, see examples/README

this being said I cannot be 100% positive that this has worked perfectly in the past, given that the failure can easily go unnoticed - in our case at least (this is because the README on purpose uses a theme very close to the default)

---
so yes, this clearly has to do with jupyterhub acting less transparently than when running off a local disk; I'm not quite sure how to handle that properly

* on the one hand, the way we search for custom css is questionable; and as result, among other things, the error messages that show up in the console are confusing, because the custom files are only optional so their absence should not trigger an error

* on the other hand, I am not aware of a mechanism - api or other - that would let us do something equivalent; our requirements would to keep it as simple (i.e. staying away from jupyter abstruse config system), in a less confusing way (no more red herring), and that would be guaranteed to work in a hosted environment like jupyterhub 

if such a mechanism was available rise would clearly benefit from using it :)

in the meanwhile,  it's be interesting to ask jupyterhub on their thoughts about that behaviour

damianavila:
Let's ping some JupyterHub/Binder devs here, maybe they have some quick feedback for us.

cc @minrk @choldgraf @betatim

betatim:
Who/how is the request for `https://hub.gke.mybinder.org/hub/files/examples/rise.css` generated? It looks like who ever is doing that generates an absolute URL instead of a relative one (and assumes that `/hub/` is a good starting value).

If possible you need to generate a URL relative to the notebook or do some (very clever) stuff looking at the URL you are being served from to figure out what the absolute URL should be.

sbrunk:
Thanks @parmentelat for the explanations.

[...]

PR #482, opened 2019-06-28
load css through a relative url

parmentelat:
follow up to #479

### DIFF
--- a/rise/static/main.js
+++ b/rise/static/main.js
@@ -483,20 +483,15 @@ define([
      * should be redefinable in the config
      */
     if (window.location.pathname.endsWith('.ipynb')) {
-      // typically examples/README.ipynb
-      let path = Jupyter.notebook.notebook_path;
-      // typically README.ipynb
       let name = Jupyter.notebook.notebook_name;
       // asscosiated css
       let name_css = name.replace(".ipynb", ".css");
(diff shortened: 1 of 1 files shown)


sbrunk:
Did a quick test using your branch locally and on binder and it loads `rise.css` correctly in both cases. 

It does not find `reveal.css` but I think that's unrelated (also occurs on master for me).

parmentelat:
thanks for the report; but to clarify, with reveal.css nothing works, right ?

I'll try to release a new dev release with this pr as soon as Damian gives his go

sbrunk:
Yes, it tries to load reveal.css, but gets a 404 which completely breaks the layout.

Note that I did not use development mode for testing this but installed RISE via pip instead like this:
```bash
pip install https://github.com/damianavila/RISE/archive/relative-url.zip
```
Not sure if this can make a difference here.

damianavila:
> I'll try to release a new dev release with this pr as soon as Damian gives his go

Feel free to do so!

damianavila:
> Note that I did not use development mode for testing this but installed RISE via pip instead like this:

That's probably the cause.

parmentelat:
Hi
I did release 5.6.0.dev3 earlier today, just had not yet got around to announcing it

PR #487, opened 2019-07-01
Use data-base-url to determine the Jupyter server root

jaanos:
Currently, RISE looks for the CSS stylesheets with a path derived from the relative path to the notebook. This works fine when running Jupyter locally - however, problems arise when Jupyter is not running on the server root (say, when running on Binder). This patch obtains the appropriate path from the `data-base-url` attribute of `<body>`, allowing CSS stylesheets to work on Binder as well.

### DIFF
--- a/rise/static/main.js
+++ b/rise/static/main.js
@@ -490,7 +490,10 @@ define([
       // asscosiated css
       let name_css = name.replace(".ipynb", ".css");
       // typically /files/examples/
-      let prefix = `/files/${path.replace(name, '')}`;
+      if (typeof($('body').attr("data-base-url")) == 'undefined') {
+          $('body').attr("data-base-url", '/')
+      }
+      let prefix = $('body').attr("data-base-url") + `files/${path.replace(name, '')}`;
       // Attempt to load rise.css
       $('head').append(
           `<link rel="stylesheet" href="${prefix}rise.css" id="rise-custom-css" />`);


jaanos:
I see now that #482 solves the same problem, but with a different approach. I'll keep this PR open for now.

parmentelat:
Hey jaanos

as a matter of fact that was the very purpose of PR#482 already; although the approaches are different, in that we'd had from the core jupyter team an advice to go for relative URLs - see betatim's input in #479 

any additional feedback on the relative merits of these 2 approaches are welcome

jaanos:
Thanks for the feedback. Actually, I also think that using relative paths is better when possible - especially as the current code obtains the path by substituting all occurences of the notebook name (which will, of course, give wrong results if it occurs multiple times). But maybe this approach could be useful elsewhere (for example, if per-project stylesheets are to be supported).

parmentelat:
> But maybe this approach could be useful elsewhere (for example, if per-project stylesheets are to be supported

true :)

damianavila:
> true :)

OK, maybe we can fix the conflicts and adopt this solution... @parmentelat thoughts?

parmentelat:
@damianavila : the relative-url solution implemented in #482, as recommended by betatim, and as jaanos acknowledged himself, sounds like the simplest and the most robust approach; leveraging `data-base-url` could indeed turn in handy for more convoluted scenarios (e.g. project-wide settings), 

so bottom line: improving the settings system could be discussed in a separate thread, but as far as bug-fixing is concerned (i.e. this ticket) I'd much rather stick to #482 for now, esp. as the code is otherwise boiling apart in other areas :)

jaanos:
I see that #482 has been merged, so I'm closing this PR. Thanks for the quick resolution of the problem!

(Although, until version 5.6.0 can be installed with pip, I'll still need to patch 5.5.0 for this to work in Binder.)

damianavila:
@parmentelat you recently released 5.5.1, is that one containing the patch introduced by #482, right?

parmentelat:
no, that change is in 5.6.0.dev3 but not in 5.5.1

the latter was meant as a hotfix; I'd much rather if we issued a 5.6 proper, I've been using it for quite some time now and am happy with it; plus, with this rise-reveal and jlab business in the pipe, sticking to 5.5 will only enlarge the gap between stable and latest, which is never a good thing

damianavila:
> no, that change is in 5.6.0.dev3 but not in 5.5.1

Roger that!

> the latter was meant as a hotfix; I'd much rather if we issued a 5.6 proper, I've been using it for quite some time now and am happy with it; plus, with this rise-reveal and jlab business in the pipe, sticking to 5.5 will only enlarge the gap between stable and latest, which is never a good thing

I agree with that idea. Let's continue the discussion about the 5.6 release on the roadmap issue #491.

Issue #486, opened 2019-06-30
packaging / dependency wrt reveal.js / publish a rise-reveal npm package


parmentelat:
related to #475 and #381 

this is an attempt at summarizing ideas raised in other issues

### classic notebook

of course RISE depends on reveal.js; however it is not quite possible to use it as-is
historically, we had a rather straightforward patch of reveal.css in place; as part of moving to reveal-3.8.0, the set of patches reached a little wider scope

regardless, what we do for packaging the whole thing is to locally `npm install reveal.js`, patch what needs to be patched, and repackage that as part of the `static` area of the RISE jupyter (classic) extension

also relevant, although maybe at the border of this current discussion, is the case of reveal's plugins; we have support for 2 such plugins:

* `notes` for supporting a speaker view 
* `chalkboard` for adding free-hand drawings on top of the slideshow

at this point, IIUC but Damian please confirm/infirm, what we've done about these 2 extensions is:

* `notes`: I can see a source subdir `plugin/notes` which, in my understanding again, is a derived copy of reveal's note plugin;
  * not quite sure at this point *how this evolves when upgrading to a newer reveal* 
  * another thing to confirm is, *the sources for the notes plugin ship with reveal, right* ?

*  `chalkboard` - this one I know does not ship with `reveal`
  * we use the plugin as-is
  * so we've made a PR with the original author of the chalkboard plugin, so that the package can be published on npm (and IIUC I did the only publication myself)
  * bottom line is that RISE just needs to add that dependency and we're done as far as packaging is concerned

### jupyterlab extension

I've just only started to look into that; the matter being complicated by the fact that 1.0 just came out, plus 0.35 and 1.0 come with a (slightly) different framework.

regardless, it seems like for the sake of an easy packaging for the jlab extension, having an npm package of rise-reveal - that is to say, RISE's variant of reveal.js - that would come with all the patches, would be just right

and that package would make a lot of sense with the classic RISE as well AAMOF, even if the patches required for both variants (classic and jlab) may differ

### thoughts and discussion

* first off, is that summary accurate ?
  * and if so what about the `notes` thing ? (since chalkboard can remain as it is now)
* what's best in that case:
  * 1 or 2 git repos (for classic and jlab) ?
  * 1 or 2 npm packages (`rise-reveal-classic` and `rise-reveal-jlab`) ?
[...]

PR #490, opened 2019-07-05
Split rise reveal

parmentelat:
here's the PR that you requested in #486 

it's mostly about moving stuff around, I have no idea if the usual review tool will be convenient for you to assess that

note that the notes plugin is now patched in place inside of reveal.js through a sed script, so it has essentially gone except for `patch-notes-plugins.sh`


parmentelat:
note this change needs of course to be reflected in the docs - exclusively, I think, in the devel doc  
but I have not touched on these changes at all as of yet, I prefer to do all the changes in a single pass once everything is settled

parmentelat:
closing this PR as it no longer relevant
see branch `split-rise-reveal-rebased-feb2020` instead

damianavila:
Roger that, thanks!!

PR #534, opened 2020-02-14
Split rise reveal rebased feb2020

parmentelat:
this has the same content as #490, but rebased on top of maint-5.6


damianavila:
Perfect @parmentelat, thanks for doing this!

parmentelat:
as an side note, it occurred to me that we're not going to be able to use the same `pip install --pre` trick, since we already have 5.6.1.dev3 that is reachable through that channel
I mean, given that we're essentially trying to validate packaging, since the code is essentially unchanged
So one possible way forward would be to use test.pypi.org for testing

parmentelat:
@damianavila 
I have updated the devel doc, notably `release.md` that is now in sync with the code for 5.7 (there is now an new required step for building `rise-reveal`)

Also I have added a note on how to take advantage of `test.pypi.org` to check the packaging

and so 5.7.0.dev1 is available from `test.pypi.org` and you could easily install it from there and test is locally with

    pip install --index https://test.pypi.org/simple --upgrade --pre rise

damianavila:
Thanks for updating the release instructions for 5.7 and the additional notes about how to use test pypi.

Closing this one since it is even with master.
I will "release" a 5.7.0-dev version from master so we can install it with `--pre` tag
If time permits, I will look into dev version from conda-forge as well.

Issue #539, opened 2020-03-05
Can't configure my slides


NEGU93:
I am trying to get my slides a bit pretier, so I am following this [file](https://github.com/damianavila/RISE/blob/master/doc/customize.md). I am using a weird image as background just to test it shows correcly. 
I tried both options:

1. Going to `Edit Notebook Metadata`, the info looks like this:
```
{
  "celltoolbar": "Slideshow",
  "kernelspec": {
    "name": "python3",
    "display_name": "Python 3",
    "language": "python"
  },
  "language_info": {
    "name": "python",
    "version": "3.7.6",
    "mimetype": "text/x-python",
    "codemirror_mode": {
      "name": "ipython",
      "version": 3
    },
    "pygments_lexer": "ipython3",
    "nbconvert_exporter": "python",
    "file_extension": ".py"
  },
  "rise": {
    "width": "100%",
    "height": "100%",
    "backimage": "back.png",
    "theme": "sky",
    "scroll": true,
    "transition": "zoom",
    "start_slideshow_at": "selected"
  }
}
```
2. Using the following code:
```
#!/usr/bin/env python3
from traitlets.config.manager import BaseJSONConfigManager
from pathlib import Path
path = Path.home() / ".jupyter" / "nbconfig"
cm = BaseJSONConfigManager(config_dir=str(path))
cm.update(
    "livereveal",
    {
        # "width": "100%",
        # "height": "100%",
        "theme": "sky",
        "scroll": True,
        "backimage": "back.png",
        "transition": "zoom",
        "start_slideshow_at": "selected",
     }
)
```

Neither case seems to work for me. Any thought on why it is not working|? I am using the button `Enter/Exit RISE slideshow` to see the result and also I tried with `jupyter nbconvert *.ipynb --TagRemovePreprocessor.remove_input_tags="{'remove_input'}" --to slides` but the html does not render with the changes neither.

parmentelat:
Hi

first off, I need to warn you that the backimage/overlay settings won't work under nbconvert; that is expected, apologies; that stuff will work under `reveal.js` (i.e. under RISE) only, but as I have never used `nbconvert` to produce slides, I was unaware of that dimension at the time where I wrote these..

second, what version of rise are you using ? the "Enter/Exit RISE slideshow" button sounds like a very old thing, unless that's from the button tooltip ?
in any case, what does `pip show rise` have to say about that ?

parmentelat:
[...]

PR #552, opened 2020-07-19
Advise a browser refresh for settings to take effect

goretkin:
Related to issue #539

### DIFF
--- a/doc/customize.md
+++ b/doc/customize.md
@@ -377,6 +377,8 @@ You can edit notebook metadata as follows
 
 ![](../examples/metadata.png)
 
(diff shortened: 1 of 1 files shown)


doc/customize.md
```diff
@@ -377,6 +377,8 @@ You can edit notebook metadata as follows
 
 ![](../examples/metadata.png)
 
+If settings do not take effect, try refreshing the page of the web browser.
```
damianavila, line 380: This is usually happening because the browser cache is a pretty hungry monster, maybe we can add here a 2 lines instruction (or link) about hard-refreshing the browser?

Issue #561, opened 2020-09-10
[Feature request]: option to remove overlay buttons


prosoitos:
I don't like the help and exit button. On my local machine, I got rid of them by messing up with the `main.js` file. It's not a great solution as my edits will get overridden by the next update. But mostly, if I make a notebook live on Binder, I cannot do the same thing there.

I thought that I could solve the problem thanks to the css by making them transparent, but I can't as they don't have any css-coded colours and they are just links (so the only way to make them transparent would make all my links transparent).

It would be great if there were an option not to show those buttons.

Thanks!!!

prosoitos:
(Note, I know that one can remove them with `,` but I don't want to see them at all. Plus, people watching my presentation on Binder may not know about that command and are likely to have those buttons visible the entire time).

prosoitos:
I wouldn't mind them as much if they were very small, for instance in the bottom left corner. But I find them rather big and not particularly nice-looking (I mean no offense!!!!!! This is not meant as criticism and is my very personal, very subjective, and totally biased perception!!!). Plus, the exit one in the top left end up on top of where I have logos on my presentations.

Anyway, if an option to get rid of them entirely existed or if it were possible to customize them (size, position, colour) in the css, that would be great.

Thanks!!! :slightly_smiling_face:

parmentelat:
Hi

I just just released 5.7.0-dev5
you can install this as always with
```console
pip install --pre rise
``` 
this release has a new setting `rise.show_buttons_on_startup` that you can set to false (the buttons still show up, but go away after 2 seconds)

on a regular setup you can tweak this interactively through the nbextensions configurator in jupyter classic like here
https://rise.readthedocs.io/en/stable/customize.html#the-configurator

I'm not quite sure how to go about configuring the binder image though; it's been a while since I did this for the last time
(there are many different ways to customize notebook extensions)
feel free to come back here if you hit a wall on that one

prosoitos:
Great! Thank you @parmentelat. I'll explore and report here.

PR #567, opened 2020-10-24
Adding color support to chalkboard

nopid:
reveal.js-plugins added color support to chalkboard some time ago.

This PR moves reveal.js from 3.8.0 to 3.9.0 and imports chalkboard from the last stable v3 reveal-js.plugins
Two keys are binded to color cycling : I choose Q & S on my AZERTY keyboard but there are certainly better options.

### DIFF
--- a/rise-reveal/package.json
+++ b/rise-reveal/package.json
@@ -33,16 +33,17 @@
   ],
   "scripts": {
     "build": "for target in copy patch; do npm run $target; done",
-    "copy": "mkdir -p export; for dep in reveal.js reveal.js-chalkboard; do cp -r ./node_modules/$dep/ ./export/$dep/; done",
-    "patch": "for target in patch-reveal-css patch-notes patch-themes ; do npm run $target; done",
+    "copy": "mkdir -p export; for dep in reveal.js; do cp -r ./node_modules/$dep/ ./export/$dep/; done; cp -r ./node_modules/reveal.js-plugins/chalkboard ./export/reveal.js-chalkboard/",
+    "patch": "for target in patch-reveal-css patch-notes patch-themes patch-chalkboard ; do npm run $target; done",
     "patch-reveal-css": "sed -i.upstream '11 s_^_/*_' export/reveal.js/css/reveal.css",
     "patch-notes": "bash patch-notes-plugin.sh",
     "patch-themes": "bash patch-reveal-themes.sh",
+    "patch-chalkboard": "bash patch-chalkboard.sh",
     "clean": "rm -rf node_modules export"
   },
   "dependencies": {
-    "reveal.js": "~3.8.0",
-    "reveal.js-chalkboard": "~1.0.0"
+    "reveal.js": "~3.9.0",
+    "reveal.js-plugins": "rajgoel/reveal.js-plugins#3.9.0"
   },
   "version": "380.0.2"
 }
--- a/rise-reveal/package.json
+++ b/rise-reveal/package.json
@@ -28,12 +28,12 @@
     "url": "https://github.com/damianavila/RISE/issues"
   },
   "files": [
-      "export/reveal.js",
-      "export/reveal.js-chalkboard"
+    "export/reveal.js",
+    "export/reveal.js-chalkboard"
   ],
   "scripts": {
     "build": "for target in copy patch; do npm run $target; done",
-    "copy": "mkdir -p export; for dep in reveal.js; do cp -r ./node_modules/$dep/ ./export/$dep/; done; cp -r ./node_modules/reveal.js-plugins/chalkboard ./export/reveal.js-chalkboard/",
+    "copy": "mkdir -p export; cp -r ./node_modules/reveal.js/ ./export/reveal.js/; cp -r ./node_modules/reveal.js-plugins/chalkboard ./export/reveal.js-chalkboard/",
     "patch": "for target in patch-reveal-css patch-notes patch-themes patch-chalkboard ; do npm run $target; done",
     "patch-reveal-css": "sed -i.upstream '11 s_^_/*_' export/reveal.js/css/reveal.css",
     "patch-notes": "bash patch-notes-plugin.sh",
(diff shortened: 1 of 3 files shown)


rise-reveal/package.json
```diff
@@ -33,16 +33,17 @@
   ],
   "scripts": {
     "build": "for target in copy patch; do npm run $target; done",
-    "copy": "mkdir -p export; for dep in reveal.js reveal.js-chalkboard; do cp -r ./node_modules/$dep/ ./export/$dep/; done",
-    "patch": "for target in patch-reveal-css patch-notes patch-themes ; do npm run $target; done",
+    "copy": "mkdir -p export; for dep in reveal.js; do cp -r ./node_modules/$dep/ ./export/$dep/; done; cp -r ./node_modules/reveal.js-plugins/chalkboard ./export/reveal.js-chalkboard/",
```
parmentelat, line 36: the for loop is no longer needed now then

rise-reveal/package.json
```diff
@@ -33,16 +33,17 @@
...
     "patch-reveal-css": "sed -i.upstream '11 s_^_/*_' export/reveal.js/css/reveal.css",
     "patch-notes": "bash patch-notes-plugin.sh",
     "patch-themes": "bash patch-reveal-themes.sh",
+    "patch-chalkboard": "bash patch-chalkboard.sh",
     "clean": "rm -rf node_modules export"
   },
   "dependencies": {
-    "reveal.js": "~3.8.0",
-    "reveal.js-chalkboard": "~1.0.0"
+    "reveal.js": "~3.9.0",
```
parmentelat, line 45: looks like 3.9.2 is out as well, is it intended to specify 3.9.0 ?
the tests that I have just made brought me 3.9.2 anyways, is that expected ?

rise-reveal/package.json
```diff
@@ -33,16 +33,17 @@
...
+    "reveal.js-plugins": "rajgoel/reveal.js-plugins#3.9.0"
```
parmentelat, line 46: same comment on the version number

rise-reveal/package.json
```diff
@@ -33,16 +33,17 @@
...
   },
   "version": "380.0.2"
```
parmentelat, line 48: you'd need to change this as well; 380 was for 3.8.0

parmentelat:
hi; thanks for this PR !

I made a few comments above on the code

in addition, I am concerned because during my tests of our code there were a few features that would no longer work
like, the X and ? buttons are gone, as well as the chalkboard buttons (but on this one I'm not quite sure if I was looking at a slideshow that had them enabled, so..) and the 't' keyboard shortcut no longer gives me the notes (but this one might have been broken earlier)

so, last time we upgraded reveal.js, we had a few more tweaks to do around.. did you test this version yourself ?
now, it might also very well be that I screwed up when installing that devel version

---
more general question out of curiosity: what about release 4.x ? have you invested any time looking into that ? 

thanks again !

parmentelat:
**important edit**

I just realized I was running my tests with a previous feature (#561) enabled as a leftover of my environment, that is why I was not getting the buttons !!

so it seems that only the 't' keyboard shortcut is not working, which again might very well be an earlier regression

nopid:
Hi, thanks for your comments. I will check and fix the version numbers.

I did it the other way around: made a quick hack fix to add color in a presentation two days ago and tried to propose a PR. The 't' shortcut was definitely working during my talk so maybe when I cleaned the code I forgot something.

The git of the talk is right here (with the .tar.gz hack :P) : https://gogit.univ-orleans.fr/lifo/no/quantum/

parmentelat:
> The 't' shortcut was definitely working during my talk so maybe when I cleaned the code I forgot something.

or maybe I'm the one who has something altered in my environment; I don't work on this often enough…

besides, I'm still toying around with it; the multiple colours on the chalkboard are cooool ! THANK YOU :)
I haven't seen the 2 new keystrokes documented in the ? help though, would make sense to add them there

parmentelat:
---
I just released the master branch as official 5.7.0; this is something that was long overdue anyway

with that in place, I'll be able to publish your pr once merged as 5.7.1-dev0 so it will be available through 
`pip install --pre rise`

parmentelat:
asking again in case you missed it :

> more general question out of curiosity: what about release 4.x ? have you invested any time looking into that ?

rise-reveal/package.json
nopid: 3.9.2 is fine, I'll fix package.json

rise-reveal/package.json
nopid: for reval.js-plugins the tag really is 3.9.0 on the git repository (no npm package so we fetch from git directly)

rise-reveal/package.json
nopid: going for 390.0.1 then

nopid:
I haven't looked at 4.x right now. I ported color while preparing a talk. No idea if it breaks something else.

nopid:
About the notes being broken, it is documented into your commit 987a9bfa18c08c888e3bde018fc5e4fbbe45f238

nopid:
Ok, I fixed the notes issue : now pressing 't' brings the notes window.

Two features than can be disabled by default : 
 - the X and ? buttons are only displayed for 2s before they hide (and , bring them back), it might be short for a new user ;
 - the chalkboard annotations are readOnly once you move to another slide, this can be changed in the configuration of the plugin, I am not sure it should be the default behavior for RISE

parmentelat:
> the X and ? buttons are only displayed for 2s before they hide (and , bring them back), it might be short for a new user ;

the intention was to keep the buttons by default, and to let people opt out through configuration (hence the relatively short duration)
are you saying that hiding the buttons is the default behaviour ?  if so, we have a bug

parmentelat:
thanks for fixing the notes, and for refreshing my memories; as I said I don't work on this often enough;)

nopid:
Oh I see, the 2s timeout was added in 784e7d234e13f6344289be799bb8af4518848a8f : https://github.com/damianavila/RISE/blob/67158fc2804e64626efd1ebc857159459e9cafa6/classic/rise/static/main.js#L659-L662
The behavior I observe is the buttons being there in the first place and being toggled (and thus disappearing) after 2s.

parmentelat:
looks good to me, merged

I issued a 5.7.0-dev0 that contains this PR as well as a fix for the buttons thingy, which was indeed broken

thanks for all, that is very helpful and much appreciated

nopid:
Thank you for the very reactive merge!

damianavila:
This is super @nopid! Thanks!

PR #590, opened 2021-03-02
Update list of reveal.js config options

moble:
[These lines](https://github.com/damianavila/RISE/blob/ebfcb5760823787e632b4a0c19681829474d1710/classic/rise/static/main.js#L539-L540) suggest that more options are actually passed through.  I took the descriptions from the reveal documentation.  (Note that the website is currently down, so I just looked through the [src](https://github.com/reveal/revealjs.com/blob/master/src/config.md).)

### DIFF
--- a/doc/customize.md
+++ b/doc/customize.md
@@ -249,7 +249,22 @@ category, RISE will pass through the following settings:
   numbers. Set to boolean `false` to turn off, [see `reveal.js`'s doc
   for more details](https://github.com/hakimel/reveal.js#slide-number)
 
-* as well as `history`.
+* `history` to push each slide change to the browser history
+
+* `width` and `height` set the "normal" size of the presentation.  The aspect
+  ratio will be preserved when the presentation is scaled to fit different
+  resolutions. Can be specified using percentage units.
+
+* `margin` is the factor of the display size that should remain empty around
+  the content
+
+* `minScale` bounds the smallest possible scale to apply to content
+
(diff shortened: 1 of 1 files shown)


damianavila:
@moble, thanks for the PR but even when some other options are passthrough, modifying some of them could be problematic (https://github.com/damianavila/RISE/blob/master/classic/rise/static/main.js#L110), this is why we do not advertise them. @parmentelat, maybe we need to prevent passing through some of these options... thoughts?

parmentelat:
hiya everythone

well, given that, as the Python motto has it, we're all consenting adults here, I'd rather leaver this open, but indeed a few words of warning about the possible negative effects of tweaking them could be quite helpful
this being said I would have a hard time trying to phrase this in any specific way...

damianavila:
> but indeed a few words of warning about the possible negative effects of tweaking them could be quite helpful

I would be OK with that!

> this being said I would have a hard time trying to phrase this in any specific way...

Me too :wink: maybe @moble has some words for us!

moble:
> Me too 😉 maybe @moble has some words for us!

I'm afraid I'm more lost than ever, now.  I would appreciate some direction on which options will/should work.

damianavila:
> I'm afraid I'm more lost than ever, now. I would appreciate some direction on which options will/should work.

I guess we are essentially asking you to add to your PR some note that warns the user that things could be wrong if they tweak those additional parameters (the ones you added).

PR #603, opened 2021-11-02
add alternative command-line Chome pdf export to docs

lkeegan:
This can be useful when generating a pdf of a slideshow on CI

### DIFF
--- a/doc/exportpdf.md
+++ b/doc/exportpdf.md
@@ -23,6 +23,10 @@ You can export your RISE presentation to PDF using the following procedures:
   + Enable the Background graphics option.
   + Click Save.
 
+The last step can also be done from the command line:
+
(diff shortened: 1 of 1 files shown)


parmentelat:
I have not been able to test this myself
I take it the 8000 part in that command must be tweaked to match the running server, is that correct ?

lkeegan:
Yes - it's the same url as in the existing step 2, just that instead of following step 3 to save the pdf from chrome this can be done directly from the command line.

I use this here to generate pdf versions of some RISE slides: https://github.com/ssciwr/jupyter-data-exploration/blob/main/.github/workflows/publish.yml#L26

damianavila:
Thanks, @lkeegan!!

Issue #610, opened 2022-03-16
Header and footer overlapped by slide content


nikita-mo:
I was trying to create a presentation with a header and footer. However, the header and footer are apparently not recognized as content of the slide, which results in the content either overlapping with the header or footer. Here is my CSS and HTML code I input to the overlay option. 

    .header{
      text-align: left;
      display: inline-block;
      vertical-align: middle;
      line-height: 32pt;
      color: black;
      font-family: calibri;
      border-bottom: 2.25pt solid #4f81bd;
      width: 85%;
      right: 0;
      left: 0;
      position: fixed;
      margin: auto;
      padding: 30px 30px 5px 30px;
    }

    .footer {
      margin: auto;
      position: fixed;
      right: 0;
      left: 0;
      bottom: 0;
      width: 85%;
      border-top: 2.25pt solid #f79646;
      color: #7F7F7F;
      font-family: calibri;
      font-size: 12pt;
      text-align: center;
    }

    .rise-enabled #notebook-container {
      margin-top: 140px;
    }

    <div class="header"><img src="logo-links.png" width='150px' padding='20px' align='left'/> <hf1>Basic methods for energy modelling and optimization in Python</hf1><img src="logo-rechts.png" width='100px' padding='20px' align='right'/><br/> <hf2>Numpy & Pandas</hf2></div><div class="footer"><p>Zentrum für nachhaltige Energiesysteme (ZNES)</p></div>

So far I've managed to add a margin on top which solves the overlapping with the header. For the footer on the other hand I couldn't find the right element to change. I added the scroll option, but the content is still stretching to the bottom of the slide, overlapping with the footer. 

I tried looking for the right container to change in my css file but nothing worked. Is there a way to add margin to the bottom as well or set the height of the content of the slide so that there is no overlapping anymore, but I can still scroll normally? 

Any help is appreciated! 

Thank you

Nikita

parmentelat:
Hi
in the meanwhile, you may give this a try if you do this
```
pip install -U --pre rise
```

and with that version in place, you will have the ability to set the height of the scroll area: 
add in your `rise.css` or other `mynotbeook.css` something along these lines:
```
/* make sure to pick a selector that is strictly 
    more selective than just .rise-scroll 
    because that's what main.css uses already
*/
section.rise-scroll {
    margin-top: 100px;  /* a bit more than should match the header size */
    height: 85vh;           /* to leave some space below for the footer */
}
```

[...]

PR #611, opened 2022-03-17
attempt to fix issue #610

parmentelat:
when scroll is selected, no longer hard-wire a height for the scroll area
instead, set a class on those elements, that are default-styled in main.less
but can be easily overridden by users in their CSS

this commit comes with a new example notebook named `header-footer-scroll` that illustrates how to leverage this change

### DIFF
--- a/examples/header-footer-scroll.ipynb
+++ b/examples/header-footer-scroll.ipynb
@@ -97,19 +97,6 @@
    "footer": "<h3>world</h3>",
    "header": "<h1>Hello</h1>",
    "scroll": true
-  },
-  "toc": {
-   "base_numbering": 1,
-   "nav_menu": {},
-   "number_sections": false,
-   "sideBar": true,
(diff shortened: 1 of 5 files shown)


examples/header-footer-scroll.ipynb
```diff
@@ -0,0 +1,117 @@
...
+   "version": "3.9.7"
+  },
+  "rise": {
+   "autolaunch": true,
+   "backimage": "mybackimage.png",
+   "footer": "<h3>world</h3>",
+   "header": "<h1>Hello</h1>",
+   "scroll": true
+  },
+  "toc": {
```
damianavila, line 101: I would prefer an example without toc metadata (to prevent confusion), but if getting rid of that take you more than 5 min, just keep it as is...

damianavila (approved):
Minor comment, otherwise LGTM.

PR #624, opened 2022-10-28
Add center to documentation

pamelafox:
RISE does actually support passing through "center" to reveal.js but the docs don't say it's one of the supported options. This PR adds it to the docs.


parmentelat:
thanks for the clarification

parmentelat:
turns out the build at readthedocs is broken, and has been so for a little while
let me try and get it back on track..

parmentelat:
and for the record we seem to be running into something that smells like this
https://github.com/readthedocs/readthedocs.org/issues/9037

parmentelat:
ok so readthedocs is working again

Issue #625, opened 2022-10-31
Weird layout when commands output large amounts of text


multimeric:
I'm using the bash kernel, but I assume this isn't specific to that kernel.

If I create a cell that has tons of output, like `grep --help` here, the cell is initially positioned correctly:

![image](https://user-images.githubusercontent.com/5019367/198946014-f0683928-c66d-418d-8226-ff3a64ad81ab.png)

Then, when I run the cell, the output is now present, but the cell is still positioned in the center of the screen instead of the top:
![image](https://user-images.githubusercontent.com/5019367/198946036-4965ba90-73ee-4a14-90e0-6b082722c3ad.png)

When I go forward a slide and then back, this resolves itself:
![image](https://user-images.githubusercontent.com/5019367/198946050-f73a3da5-8276-4d5c-90b6-54c087b34c12.png)

Here's a notebook that can be used to reproduce this. Note that *this uses the Bash kernel*:

[rise_bug.ipynb.txt](https://github.com/damianavila/RISE/files/9898581/rise_bug.ipynb.txt)

multimeric:
This seems to be related to the `top` property of the `<section id="slide-0-0">` element. It must be being calculated wrongly. In my example, it's initially set to `538.3px`, which correctly puts it in the middle of the page. Then when I run the cell, it changes to `509.8px`, so it does update, but not very much. Then finally when I re-open the slideshow, it has `top: 0px`. I wonder if, because this output is printed to stdout progressively by the command, Jupyter calculates the new location of the cell before all of the output has all been produced, and then stops updating?

Does anyone have any idea what part of RISE calculates this offset, and where to amend this behaviour?

parmentelat:
hiya

I remember having struggled with this behaviour at the time
IIRC its root cause is some sort of race condition between the various engines at work in the system, namely the execution of the cell by jupyter, its rendering by codemirror, and reveal's own internal machinery
I can't be more accurate at this point because it was all quite a while ago, but I also remember having added a configurable timer that you can tweak in the configuration area of RISE
you may want to try and enlarge this a bit

<img width="626" alt="image" src="https://user-images.githubusercontent.com/4519227/199326868-d078c059-1cde-4c52-aae7-dcde856e6c57.png">

multimeric:
[...]

PR #626, opened 2022-11-02
Fix sync_timeout parameter

multimeric:
Currently the `sync_timeout` argument cannot be properly configured because it is mistakenly called `aync_timeout` in the config.


multimeric:
Relates to, but does not entirely resolve #625.

parmentelat:
thanks for catching that :)

